PR #22, opened 2023-03-21
Implement fund/release methods

arrusev:
* implement `release` and `fund` methods
* implement `down` function to `SubnetIDHelper` lib
* add `AccountHelper` lib to check if a given address is a multi-sig or EOA
* add `CrossMsgHelper` lib to build `fund` and `release` message struct
* add `FilAddress` module


src/lib/SubnetIDHelper.sol
```diff
@@ -58,4 +58,41 @@ library SubnetIDHelper {
...
+        SubnetID calldata subnet2
+    ) public pure returns (SubnetID memory) {
+        if (subnet1.route.length <= subnet2.route.length) {
+            return SubnetID({route: new address[](0)});
+        }
+
+        uint i = 0;
+        while (
+            i < subnet1.route.length &&
+            i < subnet2.route.length &&
```
mikirov, line 73: if we get to this point aren't we guaranteed that `subnet1.route.length <= subnet2.route.length`? shouldn't one check be enough?

src/lib/SubnetIDHelper.sol
```diff
@@ -58,4 +58,41 @@ library SubnetIDHelper {
...
+        address[] memory route = new address[](i + 1);
+
+        for (uint j = 0; j < i; ) {
+            route[j] = subnet1.route[j];
+            unchecked {
+                j++;
+            }
+        }
+
+        if (i >= subnet2.route.length) route[i] = subnet1.route[i];
```
mikirov, line 94: why do we care if the index is more than the length of the second subnet when we are assigning it to an element from the first one? Don't we want that to be the usual case? Isn't it better to check if we haven't reached the end of the first subnet instead?

arrusev: Both checks would behave in the same way, since we know that subnet1 should be greater than subnet2. But i can agree that if we check for subnet1's length it would be easier to read

src/lib/SubnetIDHelper.sol
arrusev: Yup, we can go with `i < subnet2.route.length` check only

src/Gateway.sol
```diff
@@ -254,19 +311,98 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        bytes32 prevMsgsHash = checkpoint.data.crossMsgs.msgsHash;
+        bytes32 newMsgsHash = prevMsgsHash;
+
+        if (
+            checkpoint.hasCrossMsgMeta() && crossMsgRegistry[epoch].length == 0
+        ) {
+            require(prevMsgsHash == EMPTY_HASH, "no msgmeta found for cid");
+        }
+
+        if (crossMsgExistInRegistry[epoch][crossMsgHash] == false) {
```
mikirov, line 366: Amazing approach :)

src/Gateway.sol
```diff
@@ -71,19 +82,65 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+    // TODO: extract msgsHash from the real CID
+    function getCrossMsgsLength(
+        bytes32 msgsHash
+    ) external view returns (uint256) {
+        int64 epoch = crossMsgEpochRegistry[msgsHash];
+        return crossMsgRegistry[epoch].length;
+    }
+
+    // TODO: extract msgsHash from the real CID
+    function getCrossMsg(
```
mikirov, line 122: Do we need the explicit getter if the mapping is public?

mikirov (approved):

src/Gateway.sol
arrusev: If we receive a real CID from outside, then we'll have to transform it into bytes32 and in that case we'll need that getter.

src/Gateway.sol
```diff
@@ -254,19 +311,98 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+            distributeRewards(msg.sender, fee);
         }
     }
 
+    function fund(SubnetID calldata subnetId) external payable signableOnly {
+        require(msg.value > crossMsgFee, "not enough gas to pay cross-message");
+
+        CrossMsg memory crossMsg = CrossMsgHelper.createFundMsg(
+            subnetId,
+            msg.sender,
```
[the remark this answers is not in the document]
arrusev: I think we don't need to, because the `msg.sender` will always be an eth address. We should convert it only when it is passed from outside as a function argument

test/AccountHelper.t.sol
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    function test_IsAccount_Fails_BlsAccount() public view {
+        require(BLS_ADDREESS.isAccount() == false);
+    }
+
+    function test_IsAccount_Works_ContractConstructor() public {
+        DummyContract dc = new DummyContract();
+
+        require(dc.isAccount() == true);
```
mikirov, line 36: This test is great at showing that, if we consider an "account" everything without bytecode, the check can be bypassed by calling SA/GW methods from the contract constructor. It seems like the filecoin team is aware of this issue, shaw we ask them if we should proceed with this implementation or there are any other telling features of EOAs in FEVM?

test/CheckpointHelper.t.sol
```diff
@@ -22,7 +22,7 @@ contract CheckpointHelperTest is Test {
         checkpoint.data.epoch = 10;
 
         // checkpoint with epoch = 10
-        bytes32 expected = 0xe13e6ce468f1108376290016582facd126f48f28ef238d0551f5b7a98d8ee334;
+        bytes32 expected = 0xeb38f3be9a69f7b48ad5fecc6e052d96a0c77c986204d4c96c556f22d472cb08;
```
mikirov, line 25: Personally I would rather have the hash computed runtime so i can look into the code and have more certainty what is the actual input to get that hash

test/CrossMsgHelper.t.sol
```diff
@@ -0,0 +1,166 @@
...
+    function test_ToHash_Works_EmptyCrossMsgs() public view {
+        require(crossMsgs.toHash() == EMPTY_CROSS_MSGS_HASH);
+    }
+
+    function test_ToHash_Works_NonEmptyCrossMsgs() public {
+        crossMsg.message.nonce = 1;
+        crossMsgs.push(crossMsg);
+
+        // hash of cross msg array containing 1 element with nonce = 1
+        bytes32 expectedHash = 0xea420f08de853f9408624b29bc7b4ffea5db75e5cfb1865eda8b8326a6e264d6;
```
mikirov, line 65: Can we please expand this statement?

src/constants/Constants.sol
```diff
@@ -10,4 +10,6 @@ address constant STORAGE_MARKET_ACTOR = address(5);
 address constant VERIFIED_REGISTRY_ACTOR = address(6);
 address constant DATACAP_TOKEN_ACTOR = address(7);
 address constant BURNT_FUNDS_ACTOR = address(99);
-uint256 constant CROSS_MSG_FEE = 100 * 10**9;
+address constant MULTISIG_ACTOR = address(98);
```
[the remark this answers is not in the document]
arrusev: It is still not very clear how to determine if the caller is a multi-sig or not, so lets leave it as is for now

src/Gateway.sol
```diff
@@ -254,19 +312,98 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+
+        // commit top-down message.
+        (bool exist, Subnet storage subnet) = getSubnet(
+            crossMsg.message.to.subnetId.down(networkName)
+        );
+
+        require(exist, "couldn't compute the next subnet in route");
+
+        crossMsg.message.nonce = subnet.nonce;
+        subnet.nonce += 1;
```
bamzedev, line 336: Shouldn't we make update these after `distributeRewards` ?

src/Gateway.sol
```diff
@@ -254,19 +312,98 @@ contract Gateway is IGateway, ReentrancyGuard {
...
-            payable(msg.sender).functionCallWithValue(
-                abi.encodeWithSignature("reward()"),
-                fee
-            );
+            distributeRewards(msg.sender, fee);
         }
     }
 
+    function fund(SubnetID calldata subnetId) external payable signableOnly {
+        require(msg.value > crossMsgFee, "not enough gas to pay cross-message");
```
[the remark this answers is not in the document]
arrusev: No, you can't fund a subnet with `0` amount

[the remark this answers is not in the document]
bamzedev: ah, yes

src/lib/SubnetIDHelper.sol
```diff
@@ -58,4 +58,40 @@ library SubnetIDHelper {
...
+        address[] memory route = new address[](i + 1);
+
+        for (uint j = 0; j < i; ) {
+            route[j] = subnet1.route[j];
+            unchecked {
+                j++;
+            }
+        }
+
+        if (i < subnet1.route.length) route[i] = subnet1.route[i];
```
2dpetkov, line 93: We have 
- `subnet2.route.length` < `subnet1.route.length` (first "if")
- `i` <= `subnet2.route.length` (after the "while")

So `i` will always be < `subnet1.route.length`, no? The "for" can be `j <= i`

2dpetkov (approved):
Minor comments added

src/lib/SubnetIDHelper.sol
arrusev: You're right, the last `if` statement is redundant

src/lib/AccountHelper.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.7;
+
+import "../constants/Constants.sol";
+
+/// @title Helper library for checking account type
+/// @author LimeChain team
+library AccountHelper {
+    bytes32 constant ACCOUNT_HASH =
```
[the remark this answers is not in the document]
arrusev: done

mikirov (approved):

src/Gateway.sol
```diff
@@ -254,19 +312,98 @@ contract Gateway is IGateway, ReentrancyGuard {
...
         if (fee > 0) {
-            payable(msg.sender).functionCallWithValue(
-                abi.encodeWithSignature("reward()"),
-                fee
-            );
+            distributeRewards(msg.sender, fee);
         }
     }
 
+    function fund(SubnetID calldata subnetId) external payable signableOnly {
```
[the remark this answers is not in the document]
arrusev: The caller of fund/release could be EOA only(for now), so i would say it is safe to be like that.
Once we have the logic to determine if the caller is a multi-sig, then we can rethink the need for a non-reentrant guard

2dpetkov (approved):

arrusev:
* main
  * **PR #22** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/22" target="_blank">Graphite</a>  👈
    * **PR #30** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/30" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/22?utm_source=stack-comment).

src/constants/Constants.sol
```diff
@@ -10,4 +10,7 @@ address constant STORAGE_MARKET_ACTOR = address(5);
...
```
adlrocha, line 13: This is not correct, let's remove this constant. I don´t  think we can check accurately if the caller is a multisig address.

src/lib/AccountHelper.sol
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        return
+            size == 0 &&
+            ADDRESS_CODEHASH == _address.codehash &&
+            ADDRESS_CODEHASH == keccak256(_address.code);
+    }
+
+    function isMultisig(address _address) internal pure returns (bool) {
+        return _address == MULTISIG_ACTOR;
```
adlrocha, line 23: I don't think this check is correct, we could maybe remove from fund/release the check for the multisig address. In any case, I don't expect users to use this through multisigs (at least initially)

src/Gateway.sol
```diff
@@ -254,19 +312,98 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        subnet.topDownMsgs.push(crossMsg);
+
+        distributeRewards(subnetId.getActor(), crossMsgFee);
+    }
+
+    function release() external payable signableOnly {
+        require(msg.value > crossMsgFee, "not enough gas to pay cross-message");
+
+        (, int64 epoch, Checkpoint storage checkpoint) = checkpoints
+            .getCheckpointPerEpoch(block.timestamp, checkPeriod);
```
adlrocha, line 347: Just a sanity-check, here we get the checkpoint for the next epoch that is being populated, right?

adlrocha (approved):

src/Gateway.sol
arrusev: It will get the checkpoint that is available for the current epoch.

eg.
If a new checkpoint is created every 10 blocks( 1 epoch = 10 blocks)

`getCheckpointPerEpoch(5, 10)` will return the checkpoint for epoch 0 and so for all blocks between 0-9
`getCheckpointPerEpoch(12, 10)` will return the checkpoint for epoch 1 and so for all blocks between 10-19

That is the expected behaviour, right?

adlrocha: Yep, this is perfect. Thanks!

mikirov (approved):

PR #23, opened 2023-03-21
Feature/cross msg

mikirov:


src/lib/SubnetIDHelper.sol
```diff
@@ -3,59 +3,160 @@ pragma solidity ^0.8.7;
...
+            if (subnet1.route[i] != subnet2.route[i]) return false;
+            unchecked {
+                ++i;
+            }
+        }
+
+        return true;
+    }
+
+    function commonParent(
```
arrusev, line 96: The logic seems to be working but could you add all the test cases from the ref implementation just to be sure we haven't missed something

src/SubnetActor.sol
```diff
@@ -222,7 +228,7 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
             }
         }
 
-        bool hasMajority = sum > (totalStake * 2 / 3);
+        bool hasMajority = sum > totalStake * majorityPercentage / 100;
```
arrusev, line 231: Would it be better if we store the `sum` variable in the storage and just increment/decrement it when the voter's stake has changed, that way we can remove the for loop from above?

PR #27, opened 2023-03-27
feat: apply msg implementation

mikirov:


src/structs/Checkpoint.sol
```diff
@@ -36,6 +36,10 @@ struct CrossMsg {
     bool wrapped;
 }
 
+struct ApplyMsgParams {
```
arrusev, line 39: do we need a dedicated struct for ApplyMessage, since it contains a CrossMsg struct only? Can't we use just a CrossMsg function arg instead?

src/lib/CrossMsgHelper.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    function toHash(
+        CrossMsg[] memory crossMsgs
+    ) internal pure returns (bytes32) {
+        return keccak256(abi.encode(crossMsgs));
+    }
+
+    function execute(CrossMsg calldata crossMsg, bytes4 methodSelector) public returns (bool success) {
+        bytes memory params = crossMsg.wrapped ? abi.encode(ApplyMsgParams({crossMsg: crossMsg})) : crossMsg.message.params;
+        (success, ) = crossMsg.message.to.rawAddress.call{value: crossMsg.message.value}(
```
arrusev, line 73: probably we'll need to `normalize` the `rawAddress` because it is passed from outside

arrusev:
will be reopened via Graphite

PR #29, opened 2023-03-28
Feature/cross msg

mikirov:


src/Gateway.sol
```diff
@@ -354,7 +340,92 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+                burn = crossMessage.message.value > 0;
+                _commitBottomUpMsg(crossMessage);
+            }
+        } else {
+            console.log("here");
+            // top down message path
+            appliedTopDownNonce += 1;
+            _commitTopDownMsg(crossMessage);
+        }
+
```
arrusev, line 407: Just a suggestion...
The nested `if` in combination with `else` statements is a bit hard to read, so i would refactor it with something like that:

```
SubnetID memory from = crossMessage.message.from.subnetId;
IPCMsgType applyType = crossMessage.message.applyType(networkName);
bool shouldCommitBottomUp;
bool shouldCommitTopDown;

if (applyType == IPCMsgType.BottomUp) {
    require(from.route.length > 0, "error getting subnet from msg");
    shouldCommitBottomUp = to.commonParent(from).equals(networkName) == false;
}

if (shouldCommitBottomUp) {
    burn = crossMessage.message.value > 0;
    _commitBottomUpMsg(crossMessage);
}

shouldCommitTopDown = shouldCommitBottomUp == false;
appliedTopDownNonce += applyType == IPCMsgType.TopDown ? 1 : 0;

if (shouldCommitTopDown) {
    topDownFee = crossMsgFee;
    _commitTopDownMsg(crossMessage);
}
```

Also, you've missed to return `topDownFee` when the apply type is `TopDown`

Not tested but it should work :D(hopefully)

src/Gateway.sol
```diff
@@ -354,7 +340,92 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        _commitBottomUpMsg(crossMsg);
+
+        payable(BURNT_FUNDS_ACTOR).sendValue(releaseAmount);
+
+    }
+    function sendCross(
+        SubnetID memory destination,
+        CrossMsg memory crossMsg
+    ) external payable isRegistered {
```
arrusev, line 352: The `sendCross` function should be called by signable addresses only, so the modifier is missing here

src/lib/StorableMsgHelper.sol
```diff
@@ -0,0 +1,63 @@
...
+
+    function applyType(
+        StorableMsg calldata storableMsg,
+        SubnetID calldata curr
+    ) public pure returns (IPCMsgType) {
+        SubnetID memory sto = storableMsg.to.subnetId;
+        SubnetID memory sfrom = storableMsg.from.subnetId;
+        if (
+            curr.commonParent(sto).equals(sfrom.commonParent(sto)) &&
+            ipcType(storableMsg) == IPCMsgType.BottomUp
```
arrusev, line 43: The check `ipcType(storableMsg) == IPCMsgType.BottomUp` can simply be replaced with `sfrom.isBottomUp(sto)`

src/Gateway.sol
mikirov: We only set the `topDownFee` when the message is a bottom up and we've reached the common parent. Otherwise only the `appliedTopDownNonce` gets incremented. Personally I'm a fan of early returns, how does this approach look to you when it comes to readability?: 
```     function _commitCrossMessage(
        CrossMsg memory crossMessage
    ) internal returns (bool burn, uint256 topDownFee) {
        SubnetID memory to = crossMessage.message.to.subnetId;
        require(to.route.length > 0, "error getting subnet from msg");
        require(
            !crossMessage.message.to.subnetId.equals(networkName),
            "should already be committed"
        );

        SubnetID memory from = crossMessage.message.from.subnetId;
        require(from.route.length > 0, "error getting subnet from msg");
        SubnetID memory nearestCommonParent = to.commonParent(from);
        IPCMsgType applyType = crossMessage.message.applyType(networkName);
        bool isCommonParentCurrentNetwork = nearestCommonParent.equals(
            networkName
        );
        bool shouldCommitTopDown = (applyType == IPCMsgType.BottomUp &&
            isCommonParentCurrentNetwork) || applyType == IPCMsgType.TopDown;
        if (shouldCommitTopDown) {
            topDownFee = crossMsgFee;
            _commitTopDownMsg(crossMessage);
            return;
        }
        burn =
            applyType == IPCMsgType.BottomUp &&
            !isCommonParentCurrentNetwork &&
            crossMessage.message.value > 0;

        _commitBottomUpMsg(crossMessage);

        appliedTopDownNonce += applyType == IPCMsgType.TopDown ? 1 : 0;
    }
```

arrusev: It won't work. If the `applyType == IPCMsgType.TopDown` it will never reach to `appliedTopDownNonce` because of the early return in `shouldCommitTopDown`.

Soo, what is wrong with my approach?
I agree that we can use an early return in
```
if (shouldCommitBottomUp) {
      burn = crossMessage.message.value > 0; 
     _commitBottomUpMsg(crossMessage);
}
```

mikirov: You're right, good catch! Nothing wrong with it, it's simply personal developer preference whether or not to use early returns

arrusev: Yeah, btw, i also like the early returns

arrusev:
will be reopened via Graphite

PR #31, opened 2023-03-31
Implement applyMsg method

arrusev:
* add `applyMsg` method that triggers the execution of a cross-subnet message
* add `system actor` validation 
* optimize the `postbox` storage

PR #32, opened 2023-03-31
Implement propagate methods

arrusev:
* add `whitelistPropagator` method that whitelists a series of addresses as propagator of a cross net message
* add `propagate` method that propagates the stored postbox item

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -444,15 +444,15 @@ contract Gateway is IGateway, ReentrancyGuard {
     }
 
     /// @notice whitelist a series of addresses as propagator of a cross net message
-    /// @param postboxId - the id of the cross-net message
+    /// @param msgCid - the cid of the cross-net message
     /// @param owners - list of addresses to be added as owners
     function whitelistPropagator(
-        bytes32 postboxId,
+        bytes32 msgCid,
         address[] calldata owners
     ) external {
-        require(postboxHasOwner[postboxId][msg.sender], "not owner");
+        require(postboxHasOwner[msgCid][msg.sender], "not owner");
 
-        CrossMsg storage crossMsg = postbox[postboxId];
+        CrossMsg storage crossMsg = postbox[msgCid];
 
         require(crossMsg.isEmpty() == false, "postbox item does not exist");
 
@@ -460,8 +460,8 @@ contract Gateway is IGateway, ReentrancyGuard {
         for (uint256 i = 0; i < owners.length; ) {
             address owner = owners[i];
 
-            if (postboxHasOwner[postboxId][owner] == false) {
-                postboxHasOwner[postboxId][owner] = true;
+            if (postboxHasOwner[msgCid][owner] == false) {
+                postboxHasOwner[msgCid][owner] = true;
             }
             unchecked {
                 i++;
@@ -469,16 +469,16 @@ contract Gateway is IGateway, ReentrancyGuard {
         }
     }
 
-    /// @notice propagates the populated cross net message for the given id
-    /// @param postboxId - the id of the cross-net message
-    function propagate(bytes32 postboxId) external payable {
+    /// @notice propagates the populated cross net message for the given cid
+    /// @param msgCid - the cid of the cross-net message
+    function propagate(bytes32 msgCid) external payable {
         require(
             msg.value >= crossMsgFee,
             "not enough gas to pay cross-message"
         );
-        require(postboxHasOwner[postboxId][msg.sender], "not owner");
+        require(postboxHasOwner[msgCid][msg.sender], "not owner");
 
-        CrossMsg storage crossMsg = postbox[postboxId];
+        CrossMsg storage crossMsg = postbox[msgCid];
 
         require(crossMsg.isEmpty() == false, "postbox item does not exist");
 
@@ -488,7 +488,7 @@ contract Gateway is IGateway, ReentrancyGuard {
 
         _crossMsgSideEffects(crossMsg, shouldBurn, shouldDistributeRewards);
 
-        delete postbox[postboxId];
+        delete postbox[msgCid];
 
         uint256 feeReminder = msg.value - crossMsgFee;
 
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -106,13 +106,19 @@ contract Gateway is IGateway, ReentrancyGuard {
         _;
     }
 
-    modifier onlyPostboxOwner(bytes32 postboxId) {
+    modifier onlyPostboxOwner(bytes32 postboxCid) {
+        uint256 postboxId = postboxCidToId[postboxCid];
+
         require(postboxHasOwner[postboxId][msg.sender], "not owner");
         _;
     }
 
-    modifier onlyValidPostboxId(bytes32 postboxId) {
-        require(!postbox[postboxId].isEmpty(), "postbox item does not exist");
+    modifier onlyValidPostboxId(bytes32 postboxCid) {
+        uint256 postboxId = postboxCidToId[postboxCid];
+
+        PostboxItem memory postboxItem = postbox[postboxId];
+
+        require(!postboxItem.isEmpty(), "postbox item does not exist");
         _;
     }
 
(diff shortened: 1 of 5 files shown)


arrusev:
* main
  * **PR #22** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/22" target="_blank">Graphite</a> 
    * **PR #30** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/30" target="_blank">Graphite</a> 
      * **PR #31** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/31" target="_blank">Graphite</a> 
        * **PR #32** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/32" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/32?utm_source=stack-comment).

src/Gateway.sol
```diff
@@ -119,6 +118,22 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        uint256 postboxId = postboxCidToId[postboxCid];
+
+        require(postboxHasOwner[postboxId][msg.sender], "not owner");
+        _;
+    }
+
+    modifier onlyValidPostboxId(bytes32 postboxCid) {
+        uint256 postboxId = postboxCidToId[postboxCid];
+
+        PostboxItem memory postboxItem = postbox[postboxId];
```
2dpetkov, line 131: `postboxItem` can be referenced from `storage`

src/Gateway.sol
```diff
@@ -486,6 +500,63 @@ contract Gateway is IGateway, ReentrancyGuard {
         return abi.encode(postboxCid);
     }
 
+    function whitelistPropagator(
+        bytes32 postboxCid,
+        address[] calldata owners
+    ) external onlyValidPostboxId(postboxCid) onlyPostboxOwner(postboxCid) {
+        uint256 postboxId = postboxCidToId[postboxCid];
```
2dpetkov, line 507: optimisation suggestion: (both here and in `propagate`), we fetch `postboxId` 3 times and `postboxItem` twice.
We can translate `onlyPostboxOwner` and `onlyValidPostboxId` modifiers to `_checkX` functions and spare ourselves some opcodes. Or else - write the `require`'s directly in the code.

src/Gateway.sol
```diff
@@ -486,6 +500,63 @@ contract Gateway is IGateway, ReentrancyGuard {
         return abi.encode(postboxCid);
     }
 
+    function whitelistPropagator(
```
2dpetkov, line 503: please add `/// @notice`

src/Gateway.sol
```diff
@@ -486,6 +500,63 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        }
+
+        delete postboxCidToId[postboxCid];
+
+        bytes32 newPostboxCid = postboxItem.toHash();
+        postboxIdToCid[postboxId] = newPostboxCid;
+        postboxCidToId[newPostboxCid] = postboxId;
+    }
+
+    function propagate(
```
2dpetkov, line 531: please add `/// @notice`

src/Gateway.sol
```diff
@@ -486,6 +500,63 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+            postBoxItem.crossMsg,
+            shouldBurn,
+            shouldDistributeRewards
+        );
+
+        delete postbox[postboxId];
+        delete postboxIdToCid[postboxId];
+        delete postboxCidToId[postboxCid];
+
+        payable(msg.sender).sendValue(msg.value - crossMsgFee);
```
2dpetkov, line 557: is there a case when we `msg.value` will be equal to the fee? We will have `hasFee` failing then, of course, and I am wondering if that's okay.

src/interfaces/IGateway.sol
```diff
@@ -67,4 +67,12 @@ interface IGateway {
...
     /// - And updated the latest nonce applied for future checks.
     function applyMsg(CrossMsg calldata crossMsg) external returns (bytes memory);
+    
+    /// Whitelist a series of addresses as propagator of a cross net message.
+    /// This is basically adding this list of addresses to the `PostBoxItem::owners`.
+    /// Only existing owners can perform this operation.
+    function whitelistPropagator(bytes32 postboxCid, address[] calldata owners)
+        external;
+
+    function propagate(bytes32 postboxCid) external payable;
```
2dpetkov, line 77: Please add NatSpec

src/structs/Postbox.sol
```diff
@@ -2,10 +2,11 @@
 pragma solidity ^0.8.7;
 
 import "./Checkpoint.sol";
+import "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
```
2dpetkov, line 5: can be removed

src/Gateway.sol
arrusev: It's not okay, you are right. There could be a such case where the `msg.value` is equal to the `fee`. If there is a remaining fee, only then we should send it back to the sender, otherwise do nothing.

src/Gateway.sol
arrusev: we can do a combination of both approaches. 
A getter method with require in it(safe getter). wdyt?

2dpetkov: nice!

2dpetkov (approved):

src/Gateway.sol
```diff
@@ -473,6 +471,68 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        delete postboxCidToId[postboxCid];
+
+        bytes32 newPostboxCid = postboxItem.toHash();
+        postboxIdToCid[postboxId] = newPostboxCid;
+        postboxCidToId[newPostboxCid] = postboxId;
+    }
+
+    /// @notice propagates the populated cross net message for the given cid
+    /// @param postboxCid - the cid of the postbox item
+    function propagate(bytes32 postboxCid) external payable {
```
adlrocha, line 505: This Cid shouldn't be the one for the postbox item, but that of the cross-net message that wants to be propagated from the postbox.

adlrocha (changes requested):
I just realized there may be a conceptual mistake, propagate expects the CID of a cross-net message as an input, not that of the postboxItem object (maybe I misunderstood something)

src/Gateway.sol
```diff
@@ -473,6 +471,68 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+                postboxItem.owners.push(owner);
+            }
+            unchecked {
+                i++;
+            }
+        }
+
+        delete postboxCidToId[postboxCid];
+
+        bytes32 newPostboxCid = postboxItem.toHash();
```
adlrocha, line 498: Why do you need to re-compute the cid of the postbox? The postbox is used to queue cross-messages that need to be propagated further. The index used to identify an item of the postbox is through the cid of the message queued for propagation. Adding owners to the postbox item shouldn't trigger a change to the hash of the postbox message (nothing changes in the message).

src/Gateway.sol
```diff
@@ -473,6 +471,68 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+    ) external {
+        (uint256 postboxId, PostboxItem storage postboxItem) = _getSafePostboxItem(postboxCid);
+
+        // update postbox item with the new owners
+        for (uint256 i = 0; i < owners.length; ) {
+            address owner = owners[i];
+
+            if (postboxHasOwner[postboxId][owner] == false) {
+                postboxHasOwner[postboxId][owner] = true;
+                postboxItem.owners.push(owner);
```
adlrocha, line 489: Why do we need to push the owner here? Wouldn't it be enough by tracking it in `postboxHasOwner`?

adlrocha (changes requested):
LGTM apart from the thing that I mentioned in the previous PR. Unless I have misunderstood something, the postbox and postbox items don't need to be indexed by unique id, the `Cid` that we use to identify messages queued for propagation is the `Cid` of the message being queued. That means that `propagate` and `whitelistPropagator` should accept as an argument the Cid of the cross-net message that wants to be propagated further. 

Let me know if this is not clear, we can discuss sync in Slack or with quick huddle. Thanks!

src/Gateway.sol
arrusev: We were pushing it to the array, so to be able to recalculate the cid of the postbox but it's not needed anymore.

src/Gateway.sol
```diff
@@ -443,6 +443,60 @@ contract Gateway is IGateway, ReentrancyGuard {
         return abi.encode(cid);
     }
 
+    /// @notice whitelist a series of addresses as propagator of a cross net message
+    /// @param postboxId - the id of the cross-net message
+    /// @param owners - list of addresses to be added as owners
+    function whitelistPropagator(
+        bytes32 postboxId,
```
adlrocha, line 450: Maybe rename throughout the code? There is no specific ID for the postbox, and it may be confusing to use `postboxID`. The postbox is the queue of messages that need to be propagated, but the postbox in itself doesn't  have an ID. 
```suggestion
        bytes32 msgCid,
```

src/interfaces/IGateway.sol
```diff
@@ -67,4 +67,13 @@ interface IGateway {
     /// - Performs the corresponding state changes.
     /// - And updated the latest nonce applied for future checks.
     function applyMsg(CrossMsg calldata crossMsg) external returns (bytes memory);
+    
+    /// Whitelist a series of addresses as propagator of a cross net message.
+    /// This is basically adding this list of addresses to the `PostBoxItem::owners`.
+    /// Only existing owners can perform this operation.
+    function whitelistPropagator(bytes32 postboxId, address[] calldata owners)
```
adlrocha, line 74:
```suggestion
    function whitelistPropagator(bytes32 msgCid, address[] calldata owners)
```

src/interfaces/IGateway.sol
```diff
@@ -67,4 +67,13 @@ interface IGateway {
...
+        external;
+
+    /// Propagates the stored postbox item for the given cid
+    function propagate(bytes32 postboxId) external payable;
```
adlrocha, line 78:
```suggestion
    function propagate(bytes32 msgCid) external payable;
```

adlrocha (approved):
LGTM. Just a quick nit to rename the argument for the function (and maybe throughout the code, as using `postboxID` maybe a bit confusing, because there is no ID for the postbox. Apart from that, feel free to merge. Thanks for the fix.

2dpetkov (dismissed):

mikirov (dismissed):

mikirov (approved):
LGTM

PR #35, opened 2023-04-12
Checkpoint refactoring

mikirov:


src/SubnetActor.sol
```diff
@@ -77,37 +105,38 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
...
 
+    constructor(ConstructParams memory params) {
+        require(params.majorityPercentage <= 100, "majorityPercentage must be <= 100");
+        parentId = params.parentId;
+        name = params.name;
+        ipcGatewayAddr = params.ipcGatewayAddr;
+        consensus = params.consensus;
+        minValidatorStake = params.minValidatorStake < MIN_COLLATERAL_AMOUNT ? MIN_COLLATERAL_AMOUNT : params.minValidatorStake;
+        minValidators = params.minValidators;
+        bottomUpCheckPeriod = params.bottomUpCheckPeriod < DEFAULT_CHECKPOINT_PERIOD ? DEFAULT_CHECKPOINT_PERIOD : params.bottomUpCheckPeriod;
```
2dpetkov, line 130: shouldn't we set the check periods to `DEFAULT_CHECKPOINT_PERIOD` only when the constructor value is `0`?

src/Gateway.sol
```diff
@@ -42,107 +49,114 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        uint64 bottomUpCheckPeriod;
+        uint64 topDownCheckPeriod;
+        uint64 genesisEpoch;
+        uint256 msgFee;
+    }
+
+    constructor(ConstructorParams memory params) {
+        networkName = params.networkName;
+        minStake = MIN_COLLATERAL_AMOUNT;
+        bottomUpCheckPeriod = params.bottomUpCheckPeriod > DEFAULT_CHECKPOINT_PERIOD ? params.bottomUpCheckPeriod : DEFAULT_CHECKPOINT_PERIOD;
```
2dpetkov, line 126: shouldn't we set the check periods to `DEFAULT_CHECKPOINT_PERIOD` only when the constructor value is `0`?

src/Gateway.sol
```diff
@@ -153,14 +167,15 @@ contract Gateway is IGateway, ReentrancyGuard {
...
         );
 
-        (bool registered, Subnet storage subnet) = getSubnet(msg.sender);
+        SubnetID memory id = networkName.createSubnetId(msg.sender);
+        Subnet storage subnet = subnets[id.toHash()];
 
-        require(registered == false, "subnet is already registered");
+        require(_subnetExists(subnet) == false, "subnet is already registered");
 
         subnet.id = networkName.createSubnetId(msg.sender);
```
2dpetkov, line 175: We have already created an `id` above,  we shall reuse it
 (I think we have refactored this in the other stack, so this comment might be irrelevant)

src/SubnetActor.sol
mikirov: `MIN_CHECKPOINT_PERIOD` is a more fitting name. I copy pasted the name from the original implementation, however you're right, it can be misleading

src/SubnetActor.sol
```diff
@@ -80,39 +103,48 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
...
+        majorityPercentage = params.majorityPercentage;
         status = Status.Instantiated;
     }
+
     receive() external payable onlyGateway {}
 
-    function join() external payable mutateState {
+    function join() external payable mutateState signableOnly {
         require(
             msg.value > 0,
```
2dpetkov, line 149: what will happen if an address calls `join` with `msg.value` >0 and <`minValidatorStake`? Aren't we supposed to revert the transaction?

src/lib/CheckpointHelper.sol
```diff
@@ -8,43 +8,57 @@ import "../constants/Constants.sol";
 /// @title Helper library for manipulating Checkpoint struct
 /// @author LimeChain team
 library CheckpointHelper {
-    bytes32 private constant EMPTY_CROSSMSG_HASH =
+    bytes32 public constant EMPTY_TOP_DOWN_CHECKPOINT_HASH =
```
2dpetkov, line 11: I don't think we use this

src/lib/CheckpointHelper.sol
```diff
@@ -8,43 +8,57 @@ import "../constants/Constants.sol";
...
-                CrossMsgMeta({msgsHash: EMPTY_HASH, nonce: 0, value: 0, fee: 0})
+                TopDownCheckpoint({
+                    epoch: 0,
+                    topDownMsgs: new CrossMsg[](0)
+                })
             )
         );
 
-    bytes32 public constant EMPTY_CHECKPOINT_DATA_HASH =
+    bytes32 public constant EMPTY_BOTTOM_UP_CHECKPOINT_HASH =
```
2dpetkov, line 21: I don't think we use this

src/Gateway.sol
```diff
@@ -598,43 +550,87 @@ contract Gateway is IGateway, ReentrancyGuard {
...
-        crossMessage.message.nonce = subnet.nonce;
-        subnet.nonce += 1;
+        crossMessage.message.nonce = subnet.topDownNonce;
+        subnet.topDownNonce += 1;
         subnet.circSupply += crossMessage.message.value;
         subnet.topDownMsgs.push(crossMessage);
     }
 
     /// @notice commit bottomup messages for their execution in the subnet
     function _commitBottomUpMsg(CrossMsg memory crossMessage) internal {
```
2dpetkov, line 560: we are not updating `bottomUpCheckpoints` in `commitChildCheck`, so this ends up the only place where we create new bottom-up checkpoints for the storage mapping, and we do not initiate them fully.

src/lib/SubnetIDHelper.sol
```diff
@@ -146,4 +147,10 @@ library SubnetIDHelper {
         
         return SubnetID({route: route});
     }
+
+    function isBottomUp(SubnetID calldata from, SubnetID calldata to) public pure returns (bool){
```
2dpetkov, line 151: we only use this in the tests, so we might move it there (just a suggestion)

src/Gateway.sol
```diff
@@ -111,29 +112,28 @@ contract Gateway is IGateway, ReentrancyGuard {
...
     }
 
-    function getCrossMsg(
-        bytes32 msgsHash,
-        uint256 index
-    ) external view returns (CrossMsg memory) {
-        int64 epoch = crossMsgEpochRegistry[msgsHash];
-        return crossMsgRegistry[epoch][index];
+    constructor(ConstructorParams memory params) {
+        require(majorityPercentage <= 100);
```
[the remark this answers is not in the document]
mikirov: Great catch!

src/Gateway.sol
```diff
@@ -164,14 +164,16 @@ contract Gateway is IGateway, ReentrancyGuard {
             "call to register doesn't include enough funds"
         );
 
-        (bool registered, Subnet storage subnet) = _getSubnet(msg.sender);
+        SubnetID memory subnetId = networkName.createSubnetId(msg.sender);
```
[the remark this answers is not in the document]
mikirov: Not quite. We are usind the subnetId to set the subnet storage variable later. This line inside the overloaded getter is used for hashing only. If we take the `subnet.id` in this case we are (in the happy path where it's not registered) get zero instead of an id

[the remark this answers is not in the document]
2dpetkov: got it, thanks

src/SubnetActor.sol
mikirov: We will forward the money to the GW, calling `addStake` instead. The rust implementation does not revert as long as there is msg.value > 0

mikirov: Actually the variable `minValidatorStake` is misleading, it's not the min amount to become a validator but the min to activate the subnet. Perhaps a name such as `minActivationCollateral` would be more appropriate

src/Gateway.sol
```diff
@@ -672,6 +667,13 @@ contract Gateway is IGateway, ReentrancyGuard {
 
     function _applyMessages(CrossMsg[] memory crossMsgs) internal {
         for (uint i = 0; i < crossMsgs.length; ) {
+            if (i >= 1) {
```
2dpetkov, line 670: optimisation idea:
start the loop with a `prevNonce = 0` and compare `crossMsgs[i].message.nonce` to it - we will spare ourselves an `if` branch on each cycle

src/Gateway.sol
mikirov: Yes, i had optimized away (removed) the child checks and only executed the cross messages since they were never used. After the weekly yesterday we got to the conclusion that we still need them and we returned them in `commitChildCheck` :)

src/SubnetActor.sol
```diff
@@ -42,21 +48,33 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
...
     bytes public genesis;
     /// @notice number of blocks after which finality is reached
-    int64 public finalityThreshold;
-    /// @notice number of blocks between two checkpoints
-    int64 public checkPeriod;
-    /// @notice block number to corresponding checkpoint at that block
-    mapping(int64 => Checkpoint) public checkpoints;
-    /// @notice keccak256 hashed message data to set of validators who voted for the checkpoint
-    mapping(bytes32 => EnumerableSet.AddressSet) private windowChecks;
+    uint64 public finalityThreshold;
```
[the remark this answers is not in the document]
arrusev: No, it was left from the old version.
Removed.

2dpetkov (dismissed):

src/SubnetActor.sol
```diff
@@ -176,75 +206,43 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
...
+        commitVoteAmount[commitHash] += stake[msg.sender];
 
-        uint sum = 0;
-        for (uint i = 0; i < voters.length(); ) {
-            sum += stake[voters.at(i)];
-            unchecked {
-                ++i;
-            }
-        }
+        bool hasMajority = commitVoteAmount[commitHash] * 100 >
```
adlrocha, line 237: This doesn't consider the case where there is no way to reach consensus and the voting need to be reset to support new votes from all validators, right? In the refactor, the `Voting` structure considers this case. If we don't do this, the checkpoint submission for an epoch may get stuck.

src/interfaces/IGateway.sol
```diff
@@ -24,8 +24,8 @@ interface IGateway {
     /// CommitChildCheck propagates the commitment of a checkpoint from a child subnet,
     /// process the cross-messages directed to the subnet.
     function commitChildCheck(
-        Checkpoint calldata checkpoint
-    ) external returns (uint);
+        BottomUpCheckpoint calldata checkpoint
```
adlrocha, line 27: Can we rename the type here? We now have two different data structures for each of the checkpoint types (top-down and bottom-up). This applies to all the code above/below. Thanks :pray: 
```suggestion
        BottomUpCheckpoint calldata bottomupCheckpoint
```

src/interfaces/IGateway.sol
```diff
@@ -76,4 +66,9 @@ interface IGateway {
 
     /// Propagates the stored postbox item for the given cid
     function propagate(bytes32 msgCid) external payable;
+
+    function submitTopDownCheckpoint(TopDownCheckpoint memory checkpoint) external;
```
adlrocha, line 70: We should use the right type here.
```suggestion
    function submitTopDownCheckpoint(TopDownCheckpoint memory topdownCheckpoint) external;
```

src/lib/CheckpointHelper.sol
```diff
@@ -8,43 +8,15 @@ import "../constants/Constants.sol";
...
-                        value: 0,
-                        fee: 0
-                    })
-                })
-            )
-        );
-
     function toHash(
-        Checkpoint memory checkpoint
+        BottomUpCheckpoint memory checkpoint
```
adlrocha, line 12: As mentioned above, topdown and bottomup checkpoints are different structs and include different types of data.

adlrocha (changes requested):
I did first pass to the code leaving a few comments. Let me know if they make sense or I misunderstood something. Thanks :pray:

arrusev:
* main
  * **PR #35** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/35" target="_blank">Graphite</a>  👈
    * **PR #38** <a href="https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/38" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/LimeChain/filecoin-ipc-actors-fevm/35?utm_source=stack-comment).

src/interfaces/IGateway.sol
arrusev: Sure. 
Just to mention that the type is `BottomUpCheckpoint` and the `checkpoint` is just the variable name but we'll rename it for readability.

src/lib/CheckpointMappingHelper.sol
```diff
@@ -9,35 +9,22 @@ import "../structs/Checkpoint.sol";
...
-    ) public view returns (bytes32) {
-        epoch -= checkPeriod;
-        while (checkpoints[epoch].signature.length == 0 && epoch > 0) {
-            epoch -= checkPeriod;
-        }
-
-        return checkpoints[epoch].toHash();
-    }
-
     function getCheckpointPerEpoch(
```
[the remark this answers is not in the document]
arrusev: The function is used to get the next checkpoint according the current block number as it was before. I see there is `+1` added to the equation, is it because the very first epoch should start from `1` and not from `0` ?

src/SubnetActor.sol
arrusev: I think we support that case now, could you take a look

src/structs/EpochVoteSubmission.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.7;
+
+import "./Checkpoint.sol";
+
+struct EpochVoteSubmission {
+    uint256 nonce;
+    uint256 totalSubmissionWeight;
+    bytes32 mostVotedSubmission;
+    mapping(uint256 => mapping(address => bool)) submitters;
```
2dpetkov, line 10:
```suggestion
    // nonce => validator => hasSubmitted
    mapping(uint256 => mapping(address => bool)) submitters;
```

src/structs/EpochVoteSubmission.sol
```diff
@@ -0,0 +1,13 @@
...
+    mapping(uint256 => mapping(bytes32 => uint256)) submissionWeights;
```
2dpetkov, line 11:
```suggestion
    // nonce => submissionHash => weight
    mapping(uint256 => mapping(bytes32 => uint256)) submissionWeights;
```

src/structs/EpochVoteSubmission.sol
```diff
@@ -0,0 +1,13 @@
...
+    mapping(bytes32 => TopDownCheckpoint) submissions;
```
2dpetkov, line 12:
```suggestion
    // submissionHash => submission
    mapping(bytes32 => TopDownCheckpoint) submissions;
```

src/lib/EpochVoteSubmissionHelper.sol
```diff
@@ -0,0 +1,49 @@
...
+        EpochVoteSubmission storage voteSubmission,
+        TopDownCheckpoint calldata submission,
+        uint256 voteSubmissionWeight
+    ) internal {
+        bytes32 submissionHash = submission.toHash();
+
+        voteSubmission.submitters[voteSubmission.nonce][msg.sender] = true;
+        voteSubmission.totalSubmissionWeight += voteSubmissionWeight;
+        voteSubmission.submissionWeights[voteSubmission.nonce][submissionHash] += voteSubmissionWeight;
+
```
2dpetkov, line 21:
```suggestion

        // store the submission only the first time
```

src/SubnetActor.sol
```diff
@@ -176,75 +214,50 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
         IGateway(ipcGatewayAddr).kill();
     }
 
-    function submitCheckpoint(Checkpoint calldata checkpoint) external {
+    function submitCheckpoint(
```
2dpetkov, line 217: We need to account for voting here as well

2dpetkov (changes requested):

2dpetkov (approved):
Looks good to me!
@mikirov do have a critical view on the contracts, quite a lot has changed since your initial refactoring

src/lib/CheckpointMappingHelper.sol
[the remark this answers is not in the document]
adlrocha: In Rust we use this function to get the next checkpoint that needs to be populated according to the current epoch, thus the `+1`. We want the ceiling of the division between `curr_epoch` and `period`. If we are in epoch 7, and we receive a cross-net message or a child checkpoint, we need to assign these to the template of the checkpoint at epoch 10 (i.e. the next one being populated). This is why I ask that it really depends on how you are using this function. 

Regarding the initial checkpoint, the first epoch for a checkpoint is `genesis_epoch + check_period` so no checkpoint is current populated for epoch 0 if `genesis_epoch=0`.

src/Gateway.sol
```diff
@@ -347,23 +368,96 @@ contract Gateway is IGateway, ReentrancyGuard {
...
+        // submit the vote
+        bool shouldExecuteVote = _submitTopDownVote(voteSubmission, checkpoint, msg.sender, validatorWeight);
+        
+        CrossMsg[] memory topDownMsgs;
+        
+        if (shouldExecuteVote) {
+            topDownMsgs = _markMostVotedSubmissionExecuted(voteSubmission);
+        }
+
+        // there is no execution for the current submission, so get the next executable epoch from the queue
```
adlrocha, line 448: Strictly speaking, we may execute a checkpoint in the previous stage with no top-down message. This is fine, as what we don't want is to execute more than one checkpoint with cross-net message as it may lead to the block running out of gas. I suggest changing the comment to avoid confusions. 
```suggestion
        // no messages executed in the current submission, let's get the next executable epoch from the queue to see if it can be executed already
```

src/Gateway.sol
```diff
@@ -490,42 +540,37 @@ contract Gateway is IGateway, ReentrancyGuard {
...
-        ) {
-            appliedBottomUpNonce += 1;
-        } else if (
-            appliedBottomUpNonce == MAX_NONCE && storableMsg.nonce == 0
-        ) {
-            appliedBottomUpNonce = 0;
+    function _markMostVotedSubmissionExecuted(EpochVoteTopDownSubmission storage voteSubmission) internal returns(CrossMsg[] storage){
+        TopDownCheckpoint storage mostVotedSubmission = voteSubmission.submissions[voteSubmission.vote.mostVotedSubmission];
+
+        if (mostVotedSubmission.isEmpty() == false) {
```
adlrocha, line 553: Quick sanity-check: Can this ever be empty? In other words, can this function be called with an empty `mostVotedSubmissionExecuted`? I guess that if there is only one validator in the network this could be the case, right?

src/SubnetActor.sol
```diff
@@ -176,84 +215,65 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
...
-                checkpoint.data.prevHash ==
-                CheckpointHelper.EMPTY_CHECKPOINT_DATA_HASH,
-            "checkpoint data hash is not the same as prevHash"
-        );
-
-        bytes32 messageHash = checkpoint.toHash();
+        // the epoch being submitted is the next executable epoch, we perform a check to ensure
+        // the checkpoints are chained. This is an early termination check to ensure the checkpoints
+        // are actually chained.
+        if (_isNextExecutableEpoch(checkpoint.epoch)) {
```
adlrocha, line 235: I think we don't need this check, right? We accept the submission of checkpoints in the future at any moment after the refactor. The only thing we need to check is that it is for an epoch that is divisible by the checkpoint period. The chain comparison is explicitly done  when a checkpoint is executed, but not in the submission

src/SubnetActor.sol
```diff
@@ -293,45 +297,50 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
...
+    function _getMostVotedSubmission(EpochVoteBottomUpSubmission storage voteSubmission) internal view returns(BottomUpCheckpoint storage){
+        return voteSubmission.submissions[voteSubmission.vote.mostVotedSubmission];
     }
 
+    function _commitCheckpoint(EpochVoteSubmission storage vote, BottomUpCheckpoint storage checkpoint) internal returns(bool committed) {
+        if (checkpoint.isEmpty()) {
+            return false;
+        }
+
+        /// Ensures the checkpoints are chained. If not, should abort the current checkpoint.
```
adlrocha, line 331: Here is where the chained check is done, this is why we don't need it when submitting a checkpoint (as it removes the ability to submit valid checkpoints in the future)

adlrocha (approved):
LGTM. I left a comment for a slight bug in submit checkpoints when checking that checkpoints are chained, and a few sanity-check but nothing blocking. Feel free to address them and merge. Amazing job :pray:

src/Gateway.sol
arrusev: I think you're right. It shouldn't have a such case where the `mostVotedSubmission` is empty even if there is only one validator.

src/SubnetActor.sol
```diff
@@ -176,84 +215,65 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard {
...
+        BottomUpCheckpoint storage submissionToExecute;
 
-        uint sum = 0;
-        for (uint i = 0; i < voters.length(); ) {
-            sum += stake[voters.at(i)];
-            unchecked {
-                ++i;
-            }
-        }
+        if (shouldExecuteVote) {
```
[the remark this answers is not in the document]
arrusev: I would say let's leave it as is for now, otherwise, the solidity and rust implementations will differ, and could be hard to track the logic changes.
Also, we'll be able to focus on other things like deployment, test coverage and etc(finalize the whole implementation). Once we have that, we can try to refactor `submitCheckpoint` and make it different than the rust implementation.

arrusev (approved):

PR #46, opened 2023-05-25
Absolute import paths and deploy script

adlrocha:
This PR uses absolute paths for imported libraries to avoid the bytecode generated from having pointers to other contracts, that way the output of `forge build` can be used to deploy the contracts manually without requiring foundry. 

This PR also includes the `Deploy.sol` script @mikirov was working on as it may come handy for the deployment of the gateway.

### DIFF
--- a/deploy/Deployer.sol
+++ b/deploy/Deployer.sol
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.7;
+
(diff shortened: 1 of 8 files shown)


adlrocha:
I don't think this helps removing the need for library linking when deploying the contracts. Flattening the code doesn't help either, so closing this PR for now. I'll keep tinkering with the manual deployment see what we can do.

PR #67, opened 2023-06-08
Add subnet registry

cryptoAtwill:
Support subnet actor registry that deploys subnet contract from registry contract. At the same time, it also tracks the deployed subnet onchain.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5495 +1,29 @@
 {
   "name": "filecoin-ipc-actors-fevm",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "filecoin-ipc-actors-fevm",
-      "version": "1.0.0",
-      "license": "ISC",
-      "dependencies": {
-        "@openzeppelin/contracts": "^4.7.3"
-      },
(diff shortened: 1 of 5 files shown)


adlrocha (approved):

adlrocha (changes requested):
@cryptoAtwill, can we add a deployment scripts for the registry in `scripts/`, please? So we have all the scripts to deploy the IPC suite.

mikirov (changes requested):
We had a call with Alfonso, let's have a hardhat deployment script for the registry as well

arrusev (changes requested):
Since the `SubnetRegistry` and `Gateway` will be deployed by the same trusted actor, maybe it would be good if the `SubnetRegistry` has a constructor that accepts the address of the gateway, so deploying the subnet not relying on an externally passed gateway address.

adlrocha (dismissed):
@arrusev, @mikirov are we ok merging this one? Thanks!

mikirov (dismissed):

arrusev (approved):
LGTM.

adlrocha:
@arrusev, @mikirov, do you mind merging this one if it LGTY? I don´t  seem to have permissions :pray:

mikirov (approved):

PR #73, opened 2023-06-12
Feature/audit fixes

mikirov:


src/SubnetActor.sol
```diff
@@ -252,24 +257,35 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
...
         for (uint256 i = 0; i < validatorsLength;) {
-            payable(validators.at(i)).sendValue(rewardAmount);
+            accumulatedRewards[validators.at(i)] += rewardAmount;
             unchecked {
                 ++i;
             }
         }
     }
 
+    function withdraw() external signableOnly {
```
adlrocha, line 276: It would be useful to add some comments here describing what is the function and how it works (is not obvious from the name if you don't know the low level details of the protocol).

adlrocha (approved):
I think I checked all the issues and they have all been addressed successfully, but I let's wait for a review from @dnkolegov before merging to be sure.

src/Gateway.sol
```diff
@@ -231,6 +230,12 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
         if (registered == false) revert NotRegisteredSubnet();
 
         subnet.stake += msg.value;
+
```
dnkolegov, line 233: The implementations of addStake here and in the rust are not semantically equivalent.
For example, the Rust actor [may](https://github.com/consensus-shipyard/ipc-actors/blob/main/gateway/src/subnet.rs#L48-L52) move the subnet into an Inactive state, but this implementation doesn't.

src/SubnetActor.sol
```diff
@@ -162,12 +172,13 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
...
 
-        if (
-            stake[validator] >= minActivationCollateral && !validators.contains(validator)
-                && (consensus != ConsensusType.Delegated || validators.length() == 0)
-        ) {
-            validators.add(validator);
-            validatorNetAddresses[validator] = netAddr;
+        if (stake[validator] >= minActivationCollateral) {
+            if (!validators.contains(validator)) {
+                if (consensus != ConsensusType.Delegated || validators.length() == 0) {
```
dnkolegov, line 177: We are supporting Mir only - https://github.com/consensus-shipyard/ipc-actors/pull/122/files

dnkolegov (changes requested):

src/Gateway.sol
mikirov: Yes, because the threshold cannot change. You cannot increment a value and have it suddenly be less than the threshold. The assumed default state in the method is Inactive and it should only transition to Active, there is no way the other case can happen, right?

mikirov: Correct, The method only accepts unsigned integers so we are guaranteed to have positive values

adlrocha (approved):
After solving the merge conflicts I think we can merge. @dnkolegov can you give us the :green_circle:

dnkolegov (approved):

arrusev (approved):

PR #76, opened 2023-06-13
chore: remove unused consensusTypes

adlrocha:
Related: https://github.com/consensus-shipyard/ipc-actors/pull/122

//cc @cryptoatwill

### DIFF
--- a/src/SubnetActor.sol
+++ b/src/SubnetActor.sol
@@ -112,8 +112,9 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
     }
 
     modifier notKilled() {
-        if (status == Status.Terminating || status == Status.Killed)
+        if (status == Status.Terminating || status == Status.Killed) {
             revert SubnetAlreadyKilled();
+        }
         _;
     }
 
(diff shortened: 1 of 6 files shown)


src/SubnetActor.sol
```diff
@@ -112,7 +112,8 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
     }
 
     modifier notKilled() {
-        if (status == Status.Terminating || status == Status.Killed) revert SubnetAlreadyKilled();
+        if (status == Status.Terminating || status == Status.Killed)
```
mikirov, line 115: How did you get this formatting? Let's agree to unify the formatting using `forge fmt`?

mikirov (commented):
If we want to have only one consensus type what's the point in having an enum in the first place?

src/enums/ConsensusType.sol
```diff
@@ -4,10 +4,5 @@ pragma solidity 0.8.18;
 /// @title Subnet Consensus Type enum
 /// @author LimeChain team
 enum ConsensusType {
-    Delegated,
-    PoW,
-    Tendermint,
-    Mir,
-    FilecoinEC,
-    Dummy
+    Mir
```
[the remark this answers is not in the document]
adlrocha: Because we will have more in the future, but for now we rather have what we currently support (in the MVP we supported more consensus algorithms, but we want to release just with Mir initially).

src/SubnetActor.sol
adlrocha: I´ll push `forge fmt` (my IDE keeps messing with me whenever I save). Let's keep `forge fmt` as as base, yes.

mikirov (approved):

arrusev (approved):

Issue #79, opened 2023-06-14
Use genesisEpoch instead of zero-epoch in ExecutableQueueHelper's push function


dnkolegov:
Use genesisEpoch instead of zero-epoch in the ExecutableQueueHelper's push function:
https://github.com/LimeChain/filecoin-ipc-actors-fevm/blob/d247b20a4dd6ff41c6b9efad142cc05ffc2f8853/src/lib/ExecutableQueueHelper.sol#L7

arrusev:
the early return `if (epoch == 0) return;` considers that the epoch cannot be 0, since it is the default value of `uint` type in solidity.
So i think its better to assign `genesisEpoch` to the current active epoch, calculated based on the current block number instead of hard coding it to 0. When we use dynamic calculation of the epoch, it cannot be 0

PR #82, opened 2023-06-15
Valid epoch

arrusev:
* set `genesisEpoch` dynamically based on the current block number 
* add additional check for invalid epoch `checkpoint.epoch < genesisEpoch` 

issues:
#79 
#78


adlrocha (changes requested):
(Adding this as a reminder before merging) 
@cryptoAtwill (in case you need to address it before testing checkpoints), we'll need to review this one before merging. While the genesis epoch of the top-down message needs to be set to the epoch of the current block number in the parent, the genesis epoch for the bottomup checkpoint needs to be set to zero. See the Rust actors to double-check this.

adlrocha (approved):
LGTM. Should we merge it?

adlrocha (changes requested):
Oh wait, there may be an issue, I think all Votings shouldn't start from the current block, only the topDown, the bottomup should start from zero. I will have a deeper look tomorrow.

PR #84, opened 2023-06-16
List all validators

cryptoAtwill:
List all validators in a subnet contract. In the future we can add in pagination, but it should be good for now.

### DIFF
--- a/src/SubnetActor.sol
+++ b/src/SubnetActor.sol
@@ -303,6 +303,18 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
         return validators.at(index);
     }
 
+    /// @notice get all the validators in the subnet. TODO: we can introduce pagination
+    function allValidators() external view returns (address[] memory) {
+        uint256 length = validators.length();
+        address[] memory result = new address[](length);
+
+        for (uint256 i = 0; i < length; i++) {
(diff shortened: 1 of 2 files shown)


src/SubnetActor.sol
```diff
@@ -303,6 +303,18 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
         return validators.at(index);
     }
 
+    /// @notice get all the validators in the subnet. TODO: we can introduce pagination
```
adlrocha, line 306:
```suggestion
    /// @notice get all the validators in the subnet. 
    TODO: optionally add pagination
```

src/SubnetActor.sol
```diff
@@ -303,6 +303,18 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
...
+    function allValidators() external view returns (address[] memory) {
```
adlrocha, line 307: Should we call it `getValidators` if the name is not yet taken?
```suggestion
    function getValidators() external view returns (address[] memory) {
```

adlrocha (approved):
Two suggestions, but LGTM.

Out of a curiosity, why can't you return the array directly instead of iterating through it and passing it to a new `memory` array?

PR #85, opened 2023-06-16
Update gateway contract

cryptoAtwill:
Update gateway contract with the following:
- Checks if the validator has already voted in the top down checkpointing.
- Add `proof` field to `BottomUpCheckpoint` struct


adlrocha (approved):
LGTM.

PR #87, opened 2023-06-16
Changes for IPC Agent

cryptoAtwill:
Add view methods to expose states for IPC Agent.


adlrocha (approved):
LGTM. Let me know if you need any clarification with the Rust code for the integration that I shared :pray:

PR #89, opened 2023-06-19
List subnets

cryptoAtwill:
Add list all subnets in the registry.


adlrocha (changes requested):
I would delay a bit merging this one unless is blocking you. The behavior of the function is not the same as the one we have for the native actors.

cryptoAtwill:
@adlrocha No rush, this is main for debugging on local network since events are not returning results.

adlrocha:
I feel this is no longer needed, as it is superseded by https://github.com/consensus-shipyard/ipc-solidity-actors/pull/127. @cryptoAtwill feel free to re-open if you don't agree with this. Thanks.

PR #97, opened 2023-06-20
Query validator set

cryptoAtwill:
List the details of validators and support configuration number.


src/SubnetActor.sol
```diff
@@ -125,6 +125,12 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
         _;
     }
 
+    struct ValidatorDetail {
```
adlrocha, line 128:
```suggestion
    struct ValidatorInfo {
```

src/SubnetActor.sol
```diff
@@ -305,7 +311,7 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
 
     /// @notice get all the validators in the subnet.
     /// TODO: we can introduce pagination
-    function getValidators() external view returns (address[] memory) {
+    function getValidators() public view returns (address[] memory) {
```
adlrocha, line 314: Why changing `external` to `public`?

adlrocha (approved):
LGTM. Let's maybe call it `ValidatorInfo` instead of `ValidatorDetails` if you don't mind.

src/SubnetActor.sol
cryptoAtwill: it's called in the `getValidatorsInfo`.

src/SubnetActor.sol
cryptoAtwill: updated

src/SubnetActor.sol
cryptoAtwill: I changed back to external

src/SubnetActor.sol
```diff
@@ -175,6 +191,8 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
 
         if (stake[validator] >= minActivationCollateral) {
             if (!validators.contains(validator)) {
+                configurationNumber += 1;
```
adlrocha, line 194: The configuration number should also be increased when the stake for an existing validator is increased, as this changes their weight.

src/SubnetActor.sol
```diff
@@ -50,6 +50,9 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
     /// @notice Minimal number of validators required for the subnet to be able to validate new blocks.
     uint64 public immutable minValidators;
 
+    /// @notice Sequence number that uniquely identifies a validator set.
+    uint64 public configurationNumber;
```
adlrocha, line 54: It would be cleaner to include the configuration number in a struct aggregating all validator information, but I guess for now this is good enough.

adlrocha (changes requested):
We should increase the config number also when the stake of a validator that is already part of the subnet changes. And the sequence number only increases, even when a validator leaves.

src/SubnetActor.sol
```diff
@@ -205,6 +223,7 @@ contract SubnetActor is ISubnetActor, ReentrancyGuard, Voting {
         stake[msg.sender] = 0;
         totalStake -= amount;
         validators.remove(msg.sender);
+        configurationNumber -= 1;
```
adlrocha, line 226: Why are you reducing the configuration number here? The configuration number only increases, both when a validator joins and when it leaves.

src/SubnetActor.sol
cryptoAtwill: I created that struct, we can migrate to use that struct later

src/SubnetActor.sol
cryptoAtwill: typo, should be plus

cryptoAtwill: updated

adlrocha (approved):
LGTM now!

adlrocha (approved):

PR #107, opened 2023-06-22
Support f1 address

cryptoAtwill:
Support F1 address in EVM contracts so that fvm validators can submit their fvm address to evm contracts.


adlrocha:
@cryptoAtwill, I rebased `main` to include all the changes that we've been merging in master and I polluted a bit the history. Let's sync before merging so we target the right branch, etc.

PR #109, opened 2023-06-22
Check gateway address

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/101 and partially https://github.com/consensus-shipyard/ipc-solidity-actors/issues/56

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -115,6 +115,7 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
     error AlreadyRegisteredSubnet();
     error AlreadyInitialized();
     error InconsistentPrevCheckpoint();
+    error InvalidActorAddress();
     error InvalidPostboxOwner();
     error InvalidCheckpointEpoch();
     error InvalidCheckpointSource();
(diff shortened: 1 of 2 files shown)


src/Gateway.sol
```diff
@@ -755,6 +756,7 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
     /// @return found whether the subnet exists
     /// @return subnet -  the subnet struct
     function _getSubnet(address actor) internal view returns (bool found, Subnet storage subnet) {
+        if (actor == address(0)) revert InvalidActorAddress();
```
adlrocha, line 759: Just to double-check the `actor` here is the subnet actor of the subnet ID, right?

src/SubnetRegistry.sol
```diff
@@ -39,7 +42,6 @@ contract SubnetRegistry {
     function subnetAddress(SubnetID calldata _subnetId) external view returns (address subnet) {
         bytes32 subnetHash = _subnetId.toHash();
         subnet = subnets[subnetHash];
-
-        require(subnet != address(0), "Not exists");
+        if (subnet == address(0)) revert ZeroSubnetAddress();
```
adlrocha, line 45: Are we sure that `subnet` is an `address` here?

adlrocha (dismissed):
Just  a few sanity checks but LGTM

src/SubnetRegistry.sol
dnkolegov: Yes, because subnets is a mapping to addresses - https://github.com/consensus-shipyard/ipc-solidity-actors/blob/ce2ec9f2242bb935be5f5a9c33217a89b85f9c8a/src/SubnetRegistry.sol#L13

src/Gateway.sol
dnkolegov: According to the description:`actor the validator that created the subnet`

adlrocha (approved):

PR #111, opened 2023-06-22
Use the same Solidity version

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/98


adlrocha (approved):
LGTM. Are the versions of the rest of the contracts (mainly gateway and subnet actor) the same? Also, a quick sanity-check from @cryptoAtwill before merging, did we have any error by using this version of Solidity? I remember that we had some issue with FEVM not supporting yet the `PUSH0` operation.

adlrocha:
I tried to deploy the contracts over FEVM with the new compiler version and I can confirm that it works.

PR #112, opened 2023-06-22
Minor gas optimisations

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/95 and https://github.com/consensus-shipyard/ipc-solidity-actors/issues/106


adlrocha (approved):
LGTM.

PR #113, opened 2023-06-22
Use logical if statements

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/100


adlrocha (approved):
Way, cleaner. Thanks!

PR #114, opened 2023-06-22
Fix reentrancy

dnkolegov:
This PR fixes a potential vulnerability to reentrancy attacks - https://github.com/consensus-shipyard/ipc-solidity-actors/issues/94.
To eliminate a race, I just use the values of the message instead of one.

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -491,7 +491,9 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
         CrossMsg storage crossMsg = postbox[msgCid];
 
         (bool shouldBurn, bool shouldDistributeRewards) = _commitCrossMessage(crossMsg);
-
+// We must delete the message first to prevent potential re-entrancies, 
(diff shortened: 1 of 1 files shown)


adlrocha (changes requested):
I don't understand what this fixes and how we could perform a reentrancy in this case.

src/Gateway.sol
```diff
@@ -492,10 +492,12 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
 
         (bool shouldBurn, bool shouldDistributeRewards) = _commitCrossMessage(crossMsg);
```
adlrocha, line 494:
```suggestion
// We must delete the message first to prevent potential re-entrancies, 
// and as the message is deleted and we don't have a reference to the object
// anymore, we need to pull the data from the message to trigger the side-effects.
```

adlrocha (dismissed):

adlrocha (approved):

PR #117, opened 2023-06-22
Update contract dependencies

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/44 and updates the used libraries and Solidity version since the new open-zeppelin contracts require that.


cryptoAtwill (approved):
Not sure if this will impact on fevm, should not be, but not sure.

adlrocha (commented):
Let's not merge this one and the other one that bumps the solidity version until we have tested them. FEVM doesn´t  support `PUSH0` and recent versions of Solidity includes this new bytecode.

adlrocha:
In order to support this OpenZeppelin contract versions I had to upgrade to 0.8.19. I tested the deployment over FEVM and all seem to work fine.

adlrocha (approved):

PR #119, opened 2023-06-23
Run tests on Github Actions

dnkolegov:
This PR improves Github Actions flow:

- It adds prettier and solhint linters
- Adapts the code to prettier format
- Triages slither findings
- Adapats the makefile for old-school engineers

The checks will be green if the previous PRs are merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,16 @@
+# ==============================================================================
+# Deployment
+
 NETWORK ?= localnet
 
-check:
-	slither . --config-file ./slither.config.json
+deploy-ipc:
+	./ops/deploy.sh $(NETWORK)
 
-lint:
-	solhint 'src/**/*.sol'
+# ==============================================================================
+# Running security checks within the local computer
+
+slither:
+	slither . --config-file ./slither.config.json
 
 check-gateway:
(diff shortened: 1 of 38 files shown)


Makefile
```diff
@@ -1,18 +1,34 @@
...
 check-subnet:
 	docker run --rm -v $(shell pwd):/app -w /app mythril/myth:latest -v4 analyze --solc-json remappings.json ./src/SubnetActor.sol --solv 0.8.19
 
-deploy-ipc:
-	./ops/deploy.sh $(NETWORK)
+# ==============================================================================
+# Development support
+
+lint:
+	solhint 'src/**/*.sol'
```
adlrocha, line 25: I think the Limechain team was using `forge lint` as their base linter. I don't have a strong opinion here, so I am fine using `solhint`

adlrocha (approved):
LGTM. I haven't reviewed in depth the changes in the `.sol` files as they all seem `fmt`-related, but let me know if I should check them.

Makefile
dnkolegov: I could not find `forge lint`. I found only that - https://github.com/foundry-rs/foundry/issues/1970

adlrocha: That's fine, let's use solhint which is actually the one that IDEs use as default.

dnkolegov: We can add more linters if needed or replace that one if necessary later.

PR #120, opened 2023-06-26
Fix commit child check

cryptoAtwill:
Previous implementation is using `_getCurrentBottomUpCheckpoint`, which the epoch is pointing to the next epoch to submit. So what happens is when submitting checkpoint with target epoch, say 10, the `_getCurrentBottomUpCheckpoint` will fetch the `epoch` at 20. 
We should instead directly use the epoch provided in the checkpoint instead of using the next epoch.

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -297,16 +297,15 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
             }
         }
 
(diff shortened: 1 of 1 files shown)


adlrocha:
This was correct in the beginning. Here we need to get the checkpoint template for the next checkpoint epoch. Closing it for now.

PR #125, opened 2023-06-27
Update sendCross method

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/90

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -125,6 +125,8 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
     error InvalidCrossMsgDestinationSubnet();
     error InvalidCrossMsgDestinationAddress();
     error InvalidCrossMsgsSortOrder();
+    error InvalidCrossMsgFromSubnetId();
+    error InvalidCrossMsgFromRawAddress();
     error CannotSendCrossMsgToItself();
     error SubnetNotActive();
     error PostboxNotExist();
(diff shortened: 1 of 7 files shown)


src/Gateway.sol
```diff
@@ -502,10 +503,13 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
...
 
-        // we disregard the "to" of the message. the caller is the one set as the "from" of the message.
-        crossMsg.message.to.subnetId = destination;
-        crossMsg.message.from.subnetId = _networkName;
-        crossMsg.message.from.rawAddress = msg.sender;
+        // we disregard the "to" of the message. the caller is the one set as the "from" of the message
+        if (!crossMsg.message.from.subnetId.equals(_networkName)) {
+            revert InvalidCrossMsgFromSubnetId();
+        }
+        if (crossMsg.message.from.rawAddress != msg.sender) {
```
adlrocha, line 510: I don't think I agree with this check, why are we doing this? Actually, I am starting to consider that we should not do anything to the from and simply add the cross-net information so we can have an arbitrary `from`, or force the `from` to be the actor triggering the message.

adlrocha (approved):
My only concern, is the `from` check, I don't know if we should remove it completely :shrug:

src/Gateway.sol
dnkolegov: This line I derived from the original code - https://github.com/consensus-shipyard/ipc-solidity-actors/blob/main/src/Gateway.sol#L508.

adlrocha: did you maybe share the wrong link? This seems from the Solidity actors.

src/Gateway.sol
```diff
@@ -487,25 +490,29 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
...
-        // we disregard the "to" of the message. the caller is the one set as the "from" of the message.
-        crossMsg.message.to.subnetId = destination;
-        crossMsg.message.from.subnetId = _networkName;
-        crossMsg.message.from.rawAddress = msg.sender;
+        // We disregard the "to" of the message that will be verified in the _commitCrossMessage().
+        // The caller is the one set as the "from" of the message
+        if (!crossMsg.message.from.subnetId.equals(_networkName)) {
+            revert InvalidCrossMsgFromSubnetId();
+        }
+        // The can be many semantics of the (rawAddress, msg.sender) pairs.
```
adlrocha, line 508:
```suggestion
        // There can be many semantics of the (rawAddress, msg.sender) pairs.
```

adlrocha (approved):
LGTM. I think we can merge.

adlrocha (approved):

PR #127, opened 2023-06-27
Improved subnet registry

adlrocha:
This PR:
- Improves the subnet registry so that we can track the address of all of the subnet actors deployed by an owner. It also includes functions to access the latest subnet actor address deployed by a user. These will come really handy in the integration with the IPC agent, as we don´t  have to rely anymore on events to get the address of the subnet actor deployed. 
- It includes a batch of tests over the registry.

### DIFF
--- a/src/SubnetRegistry.sol
+++ b/src/SubnetRegistry.sol
@@ -8,9 +8,15 @@ import "./lib/SubnetIDHelper.sol";
 contract SubnetRegistry {
     using SubnetIDHelper for SubnetID;
 
-    /// @notice Mapping that tracks the deployed subnet actors.
+    /// @notice Mapping that tracks the deployed subnet actors per user.
     /// Key is the hash of Subnet ID, values are addresses.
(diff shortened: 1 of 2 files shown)


src/SubnetRegistry.sol
```diff
@@ -37,15 +43,25 @@ contract SubnetRegistry {
 
         SubnetID memory id = params.parentId.createSubnetId(subnetAddr);
 
-        bytes32 subnetHash = id.toHash();
-        subnets[subnetHash] = subnetAddr;
+        subnets[msg.sender][userNonces[msg.sender]] = subnetAddr;
+        userNonces[msg.sender]++;
```
dnkolegov, line 47: `++userNonces[msg.sender]` is more efficient from gas perspective if it is possible

dnkolegov (changes requested):

dnkolegov (approved):

cryptoAtwill (approved):

PR #129, opened 2023-06-28
Reduce contract sizes

dnkolegov:
This PR reduces the size of the main contracts making it possible to deploy the Gateway contract.


adlrocha (approved):
Amazing job. I feel this is just a transient solution, because as soon as we introduce new code we may grow the size enough for it not to be deployable, but at least it buys us time to think of a way to rearchitect it. Thanks!

PR #131, opened 2023-06-30
Get top down messages by nonce

cryptoAtwill:
Get top down messages also by nonce.


adlrocha (approved):
It is worth adding a quick test to check that this works fine?

PR #133, opened 2023-07-03
Change IPCAddress to use FvmAddress

cryptoAtwill:
Previously, `IPCAddress` is using `address` type. This makes sending bottom up messages from fvm to fevm impossible as fvm actors addresses are `IDActor` and fevm is using `address` type in solidity. This approach is more extendable as we are introducing more blockchains and it is aligned with FVM Address type implementation.

Changes:
- Support f4 address in `FvmAddress` struct.
- Add utility functions in `FvmAddressHelper` library to create `FVMAddress` easily and verify `FvmAddress` is indeed a valid eth delegated address.
- Update contract implementation to use `FvmAddress`
- Update tests.
- Use error code string and require so that reverts can be bubbled up in integration testing.


src/lib/CrossMsgHelper.sol
```diff
@@ -13,34 +15,18 @@ import {Address} from "openzeppelin-contracts/utils/Address.sol";
...
-        address signer,
+        FvmAddress calldata to,
         uint256 value
     ) public pure returns (CrossMsg memory) {
         return
             CrossMsg({
                 message: StorableMsg({
-                    from: IPCAddress({subnetId: subnet, rawAddress: BURNT_FUNDS_ACTOR}),
-                    to: IPCAddress({subnetId: subnet.getParentSubnet(), rawAddress: signer}),
+                    from: IPCAddress({subnetId: subnet, rawAddress: FvmAddressHelper.from(BURNT_FUNDS_ACTOR)}),
```
adlrocha, line 28: Let's fix this, and set the right from instead of a `BURNT_ADDRESS`, I don't think we need tis any more.

src/lib/CrossMsgHelper.sol
```diff
@@ -78,12 +65,15 @@ library CrossMsgHelper {
     }
 
     function isEmpty(CrossMsg memory crossMsg) internal pure returns (bool) {
-        return toHash(crossMsg) == EMPTY_CROSS_MSG;
+        return 
+            crossMsg.message.nonce == 0 &&
```
adlrocha, line 69: Could it be that we used the hash comparison before because it is cheaper in terms of gas?

src/lib/FvmAddressHelper.sol
```diff
@@ -1,14 +1,44 @@
...
 library FvmAddressHelper {
     /// f1: SECP256K1 key address, 20 byte hash of PublicKey.
     uint8 public constant SECP256K1 = 1;
     uint8 public constant PAYLOAD_HASH_LEN = 20;
 
+    /// For delegated FIL address
+    uint8 public constant DELEGATED = 4;
+    uint64 public constant EAM_ACTOR = 10;
+
+    /// @notice Checks if two fil addresses are the same
```
adlrocha, line 16: Is this comment correct?

src/lib/FvmAddressHelper.sol
```diff
@@ -1,14 +1,44 @@
...
+            payload: payload
+        });
+    }
+
+    function extractEvmAddress(FvmAddress memory fvmAddress) internal pure returns (address addr) {
+        require(fvmAddress.addrType == DELEGATED, "IPC-1");
+        
+        DelegatedAddress memory delegated = abi.decode(fvmAddress.payload, (DelegatedAddress));
+        
+        require(delegated.namespace == EAM_ACTOR, "IPC-1");
```
adlrocha, line 35: Let's revert to the convention we agreed on for errors.

test/CrossMsgHelper.t.sol
```diff
@@ -6,46 +6,41 @@ import "forge-std/console.sol";
...
 
     CrossMsg public crossMsg;
     CrossMsg[] public crossMsgs;
 
     error NoParentForSubnet();
 
-    function test_ToHash_Works_EmptyCrossMsg() public view {
-        require(crossMsg.toHash() == EMPTY_CROSS_MSG_HASH);
-    }
+    // TODO: comment off first, not sure what they do when address is replaced with FvmAddress
```
adlrocha, line 27: We should uncomment these tests and make them pass, let's discuss sync if needed

test/CrossMsgHelper.t.sol
```diff
@@ -203,14 +198,14 @@ contract CrossMsgHelperTest is Test {
         require(decodedCrossMsg.toHash() == crossMsg.toHash());
     }
 
-    function test_Execute_Fails_InvalidMethod() public {
-        vm.expectRevert(Address.FailedInnerCall.selector);
+    // function test_Execute_Fails_InvalidMethod() public {
```
adlrocha, line 201: Why do we comment this one?

test/FvmAddressHelper.t.sol
```diff
@@ -0,0 +1,20 @@
...
+    using FvmAddressHelper for FvmAddress;
+
+    function test_works() public pure {
+        address addr = 0xeC2804Dd9B992C10396b5Af176f06923d984D90e;
+        FvmAddress memory fvmAddr = FvmAddressHelper.from(addr);
+        
+        address extracted = fvmAddr.extractEvmAddress();
+        require(extracted == addr, "addresses not equal");
+    }
+}
```
adlrocha, line 20: Should we include more tests so we have full coverage of the helper?

test/Gateway.t.sol
```diff
@@ -1196,56 +1197,58 @@ contract GatewayDeploymentTest is StdInvariant, Test {
...
-                    value: CROSS_MSG_FEE + 1,
-                    nonce: 0,
-                    method: METHOD_SEND,
-                    params: new bytes(0)
-                }),
-                wrapped: true
-            })
-        );
-    }
+    // TODO: disable for now as we are using Fvm Address
```
adlrocha, line 1200: Let's uncomment this ones and fix them. They shouldn't be controversial, they may just need a few changes.

test/Gateway.t.sol
```diff
@@ -1766,39 +1769,40 @@ contract GatewayDeploymentTest is StdInvariant, Test {
...
         vm.prank(validators[0]);
-        vm.expectRevert(MessagesNotSorted.selector);
+        vm.expectRevert("IPC-4");
 
         gw.submitTopDownCheckpoint(checkpoint);
     }
 
-    function test_SubmitTopDownCheckpoint_Fails_InvalidCrossMsgDestinationAddress() public {
-        address validator = address(100);
+    // TODO: disable the below for now, as we are using Fvm Address.
```
adlrocha, line 1777: Same here, we should get this fixed.

adlrocha (changes requested):
We should fix some of the tests commented and make them pass before merging. Let me know if you need help with this. 
We should also get reverted the changes to errors to use our previous approach. 
Thanks!

src/lib/CrossMsgHelper.sol
cryptoAtwill: I dont think it would be cheaper than the current impl. Also we cannot use hash anymore because the EMPTY_CROSS_MSG is constant and `FvmAddress.from(address(0))` cannot be used in constant.

test/Gateway.t.sol
cryptoAtwill: as discussed we are moving these tests because FvmAddress struct is used, the address might not be a evm address only.

src/Gateway.sol
```diff
@@ -523,9 +520,11 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
         if (crossMsg.message.value != msg.value) {
             revert NotEnoughFunds();
         }
-        if (crossMsg.message.to.rawAddress == address(0)) {
-            revert InvalidCrossMsgDestinationAddress();
-        }
+
+        // TODO: disable the below for now, as we are using Fvm Address.
```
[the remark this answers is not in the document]
cryptoAtwill: this is removed.

src/lib/FvmAddressHelper.sol
cryptoAtwill: updated

test/FvmAddressHelper.t.sol
cryptoAtwill: removed unused functions

adlrocha (approved):

Issue #134, opened 2023-07-05
Do not enforce `owner` in postbox messages


adlrocha:
Same description as: https://github.com/consensus-shipyard/ipc-actors/issues/123

adlrocha:
Fixed in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/136

PR #136, opened 2023-07-06
Remove postbox owners

cryptoAtwill:
Remove postbox owners so that anyone can call `propagate`. Related to #134


adlrocha (approved):
LGTM

PR #138, opened 2023-07-10
Subnet actor and Gateway Diamonds

dnkolegov:
This PR implements Diamond pattern for SubnetActor and GatewayActor contracts.


src/SubnetActorDiamond.sol
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /// @notice Will run when no functions matches call data
+    fallback() external payable {
+        _fallback();
+    }
+
+    /// @notice Same as fallback but called when calldata is empty
+    /* solhint-disable no-empty-blocks */
+    receive() external payable onlyGateway {}
```
adlrocha, line 132: Why do we need the `OnlyGateway` modifier here?

src/facets/InfoFacet.sol
```diff
@@ -0,0 +1,152 @@
...
+    function getNetworkName() external view returns (SubnetID memory) {
+        return s.networkName;
+    }
+
+    function bottomUpCheckpoints(
+        uint64 e
+    )
+        external
+        view
+        returns (
```
adlrocha, line 53: Return data can't be specified as a `struct`?

src/facets/RouterFacet.sol
adlrocha, line 1: Can we call it `CrossFacet` instead of `RouterFacet`? Router may be confusing with our current name convention.

src/GatewayDiamond.sol
```diff
@@ -0,0 +1,97 @@
...
+import {LibVoting} from "./lib/LibVoting.sol";
+import {SubnetID, Subnet} from "./structs/Subnet.sol";
+import {SubnetIDHelper} from "./lib/SubnetIDHelper.sol";
+
+error FunctionNotFound(bytes4 _functionSelector);
+error InvalidMajorityPercentage();
+
+contract GatewayDiamond {
+    // solhint-disable-next-line private-vars-leading-underscore
+    AppStorage internal s;
```
adlrocha, line 16: Can we rename it to `GatewayActorStorage` for consistency? In the subnet actor we are naming it `SubnetActorStorage`. (I understand that in the EIP description they call it `AppStorage`).

src/lib/LibAppStorage.sol
```diff
@@ -0,0 +1,180 @@
...
+error SubnetNotActive();
+error PostboxNotExist();
+error MessagesNotSorted();
+error ValidatorsAndWeightsLengthMismatch();
+error ValidatorWeightIsZero();
+error NotEnoughFundsForMembership();
+error EpochAlreadyExecuted();
+error EpochNotVotable();
+
+struct AppStorage {
```
adlrocha, line 65: Rename to `GatewayActorStorage`?

src/SubnetActorDiamond.sol
```diff
@@ -0,0 +1,146 @@
...
+        uint256 minActivationCollateral;
+        uint64 minValidators;
+        uint64 bottomUpCheckPeriod;
+        uint64 topDownCheckPeriod;
+        uint8 majorityPercentage;
+        bytes genesis;
+    }
+
+    constructor(IDiamond.FacetCut[] memory _diamondCut, ConstructorParams memory params) {
+        LibDiamond.setContractOwner(msg.sender);
```
adlrocha, line 62: Setting the contract owner may not be as straightforward in the subnet actor (and the gateway). IPC is decentralized and we haven't figured governance yet. For now I guess the deployer could become the owner, but we should add a note to change this (the first that comes to mind is that validators should have a vote according to their weight to determine changes to the contract, but let's leave it for the future).

adlrocha (commented):
First pass, I think we are in a good path. I left a few comments. A few things that I am missing: 
- Update to the deployment scripts (which become really important with this new pattern). 
- Test and CI passing. 
- Integration of the deployment registry of subnet actors with the diamond implementation. We should also consider including the registry as an additional facet of the Gateway diamond, I think it would improve the UX and deployment a lot.

Let's sync at some point if needed.

src/SubnetActorDiamond.sol
dnkolegov: It is inherited from the main version https://github.com/consensus-shipyard/ipc-solidity-actors/blob/main/src/SubnetActor.sol#L212

src/facets/InfoFacet.sol
dnkolegov: Fixed.

src/facets/RouterFacet.sol
dnkolegov: Agreed on GatewayRouterFacet

src/lib/LibAppStorage.sol
dnkolegov: Done

scripts/python/get_selectors.py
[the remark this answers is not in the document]
dnkolegov: This file is used to get selectors of the faucets.
```
A diamond contains within it a mapping of function selectors to facet addresses. Functions are added/replaced/removed by modifying this mapping.
```

src/SubnetActorDiamond.sol
dnkolegov: Created an issue https://github.com/consensus-shipyard/ipc-solidity-actors/issues/139

hardhat.config.ts
```diff
@@ -60,7 +58,7 @@ task('deploy-gateway', 'Builds and deploys the Gateway contract on the selected
   const { deploy } = await lazyImport('./scripts/deploy-gateway');
   const gatewayDeployment = await deploy(deployments.libs);
 
-  console.log(gatewayDeployment);
+  console.log("Deployed Gateway Actor diamond address: ", gatewayDeployment);
```
adlrocha, line 61: I thinks this would break `deploy.sh` as we use the output of this task as the input to set the gateway address and automatically deploy in sequence all the IPC contracts. Check that you don't break it before adding this text, or modify `deploy.sh`. That being said, if we end up not using the registry in the end this is not critical.
```suggestion
  console.log(gatewayDeployment);
```

hardhat.config.ts
```diff
@@ -81,12 +79,43 @@ task('deploy-subnet', 'Builds and deploys the SubnetActor contract on the select
...
+
 task('deploy', 'Builds and deploys all contracts on the selected network', async (args, hre: HardhatRuntimeEnvironment) => {
   await hre.run('compile');
   await hre.run('deploy-libraries');
   await hre.run('deploy-gateway');
 });
 
+task('deploy-gw-diamond', 'Builds and deploys Gateway Actor diamond', async (args, hre: HardhatRuntimeEnvironment) => {
+  await hre.run('compile');
+  //await hre.run('deploy-libraries');
```
adlrocha, line 108: This is commented because the libraries needs to be deployed in a previous stage?  How are you planning to pass the libraries?

scripts/deploy-gw-diamond.ts
```diff
@@ -0,0 +1,160 @@
...
+    ]
+    console.log('Target facets: ', FacetNames)
+    // The `facetCuts` variable is the FacetCut[] that contains the functions to add during diamond deployment
+    const facetCuts = []
+
+    type Libraries = {
+        [libraryName: string]: string;
+    }
+
+    /*
```
adlrocha, line 30: Can you please adjust `deploy.sh` to trigger the end-to-end deployment of IPC for the diamons? (from deploying the libraries to deploying the gateway diamond).

scripts/deploy-libraries.ts
```diff
@@ -18,9 +18,18 @@ export async function deploy() {
     const { address: executableQueueHelperAddress } = await deployContractWithDeployer(deployer, "ExecutableQueueHelper", {}, txArgs);
     const { address: subnetIDHelperAddress } = await deployContractWithDeployer(deployer, "SubnetIDHelper", {}, txArgs);
     // nested libs
-    const { address: crossMsgHelperAddress } = await deployContractWithDeployer(deployer, "CrossMsgHelper", { "SubnetIDHelper": subnetIDHelperAddress }, txArgs);
-    const { address: storableMsgHelperAddress } = await deployContractWithDeployer(deployer, "StorableMsgHelper", { "SubnetIDHelper": subnetIDHelperAddress }, txArgs);
-    const { address: checkpointHelperAddress } = await deployContractWithDeployer(deployer, "CheckpointHelper", { "SubnetIDHelper": subnetIDHelperAddress }, txArgs);
+    const { address: crossMsgHelperAddress } = await deployContractWithDeployer(deployer, "CrossMsgHelper",
```
adlrocha, line 21: Has thin been automatically linted? It reads weird. There seem to be additional spaces and returns here and there.

src/README.md
```diff
@@ -0,0 +1,27 @@
+# IPC Diamond
+
+The IPC Solidity Actors are implemented in the Diamond pattern, but the current implementation 
+is not compatible with [EIP-2535](https://eips.ethereum.org/EIPS/eip-2535) standard.
```
adlrocha, line 4: out of a curiosity, why is not compatible?

scripts/python/get_selectors.py
```diff
@@ -0,0 +1,38 @@
+import subprocess
+import argparse
+import json
+
+
+from eth_abi import encode
+
+
+def get_selectors(args):
```
adlrocha, line 9: Do you mind adding a slight description in comments of why you need this and how you do it. I get that this is used to get the selector hex representation to feed it to the diamond, but for readers without context it may not be straightforward. Thanks!

adlrocha (approved):
I left a few comments with suggestions. I did my best to review this PR, but it is huge. Do you mind dividing into a set of chained PRs when this happens? You could for instance have divided the work into: 
- PR1: Diamond contracts implementation
- PR2 (with PR1 as head): Deployment scripts.
This would have made the review way more approachable. 

I would suggest making this a feature branch and not merge until we have figured the deployment of the subnet actor from the agent.

Also, a few questions: 
- Have you tried to deploy the diamonds successfully? 
- With diamonds, does it change the way in which contract functions are called? (We should understand for the IPC agent integration) 

A few

scripts/python/get_selectors.py
dnkolegov: Fixed.

src/README.md
dnkolegov: We can't add new facets or upgrade them.

dnkolegov: Added a comment to the readme.

scripts/deploy-libraries.ts
dnkolegov: Yes, automatically using Prettier.

hardhat.config.ts
dnkolegov: No, because I used the inlined static addresses and forgot to uncomment that.

hardhat.config.ts
dnkolegov: Fixed

dnkolegov:
> Have you tried to deploy the diamonds successfully?

Yes, I have. I have deployed them on the Calibration network. At least, the deployment scripts said that.

> With diamonds, does it change the way in which contract functions are called? (We should understand for the IPC agent integration)

Yes, it does.


Here is an example that uses `web3.js`:
```
const myUsefulFacet = new web3.eth.Contract(MyUsefulFacet.abi, diamondAddress);
```

Here is another example that uses hardhat:
```
const diamondLoupeFacet = await ethers.getContractAt('DiamondLoupeFacet', diamondAddress)

```

Here's an example of Solidity code that calls a function from a diamond:
```
string result = MyUsefulFacet(address(diamondContract)).getResult()
```

scripts/deploy-gw-diamond.ts
dnkolegov: Fixed

PR #140, opened 2023-07-12
add list subnets and fix deploy

adlrocha:
This PR: 
- Includes a `listSubnet` function in the gateway (the size have increased so to deploy the contract now we need to comment `propagate`, `setMembership` and `sendCross` which won't be supported until the next version of IPC)
- Add support for mainnet in the deploy scripts. 
- Add a new error of `MethodNotSupportedYet` for the mainnet deployment

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -876,9 +876,12 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
     function listSubnets() external view returns (Subnet[] memory) {
         uint256 size = subnetKeys.length;
         Subnet[] memory out = new Subnet[](size);
-        for (uint256 i = 0; i < size; ++i) {
+        for (uint256 i = 0; i < size; ) {
             bytes32 key = subnetKeys[i];
             out[i] = subnets[key];
+            unchecked {
+                ++i;
+            }
         }
         return out;
     }
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -469,7 +469,7 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
         }
 
         totalWeight = totalValidatorsWeight;
-        revert MethodNotSupportedYet()
+        // revert MethodNotSupportedYet();
     }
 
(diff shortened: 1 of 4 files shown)


cryptoAtwill (approved):

src/Gateway.sol
```diff
@@ -861,4 +870,15 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
         subnet = subnets[subnetId.toHash()];
         found = !subnet.id.isEmpty();
     }
+
+    /// @notice returns the list of registered subnets in IPC
+    /// @return subnet - the list of subnets
+    function listSubnets() external view returns (Subnet[] memory) {
+        Subnet[] memory out = new Subnet[](subnetKeys.length);
+        for (uint256 i = 0; i < subnetKeys.length; i++) {
```
dnkolegov, line 878: Could you cache the length and use `++i` in the unchecked section?

dnkolegov (changes requested):

src/Gateway.sol
adlrocha: Fixed, let me know if it looks good!

src/Gateway.sol
```diff
@@ -861,4 +870,16 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
         subnet = subnets[subnetId.toHash()];
         found = !subnet.id.isEmpty();
     }
+
+    /// @notice returns the list of registered subnets in IPC
+    /// @return subnet - the list of subnets
+    function listSubnets() external view returns (Subnet[] memory) {
+        uint256 size = subnetKeys.length;
+        Subnet[] memory out = new Subnet[](size);
+        for (uint256 i = 0; i < size; ++i) {
```
dnkolegov, line 879: By `unchecked section` I meant that
https://github.com/consensus-shipyard/ipc-solidity-actors/blob/fedd1d4bc7914da33533bda8bc95dcafaa010df4/src/SubnetActor.sol#L337

dnkolegov (changes requested):

dnkolegov (approved):

PR #144, opened 2023-07-20
Query the latest applied top down nonce

cryptoAtwill:
This is for issue #143 so that top down checkpoint submission for fevm can be implemented.

### DIFF
--- a/src/Gateway.sol
+++ b/src/Gateway.sol
@@ -746,6 +746,13 @@ contract Gateway is IGateway, ReentrancyGuard, Voting {
         return messages;
     }
 
+    /// @notice Get the latest applied top down nonce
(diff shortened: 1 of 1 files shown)


adlrocha (changes requested):
This is wrong, you are not interested in `appliedTopDownNonce` but the last top-down voting executed. 

From the gateway state you get [the voting](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/4d9a5a6bd8854a4840699fee2df90cdb978c0a96/src/SubnetActor.sol#L184) and get the [last executed](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/4d9a5a6bd8854a4840699fee2df90cdb978c0a96/src/Voting.sol#L24)

adlrocha (approved):

PR #153, opened 2023-07-24
Diamond registry

cryptoAtwill:
Integrate the diamond pattern into the subnet registry. The overall flow is as follows:
- `SubnetRegistry` is deployed with `SubnetGetterFacet` and `SubnetManagerFacet` selectors.
- `ipc-agent` calls `newSubnetActor` in `SubnetRegistry` with the function selectors as parameters.
- `SubnetRegistry` deploys the `getterFacet` and `managerFacet` and obtains their addresses.
- `SubnetRegistry` deploys a new `SubnetActorDiamond` with the provided constructor params and facet info.


scripts/deploy-registry.template
```diff
@@ -1,26 +1,78 @@
...
-import { getTransactionFees } from './util';
+import { ethers } from 'hardhat';
+import { deployContractWithDeployer, getTransactionFees } from './util';
 
-async function main() {
-  try {
+const { getSelectors } = require('./js/diamond.js')
 
-    await hre.run('compile');
+const GATEWAY = '0x4A04DDb64568C5721f6387a6e85FE1585cF7AB9e';
```
adlrocha, line 6: I guess this is for debugging purposes, right? Remove this one?

scripts/deploy-registry.template
```diff
@@ -1,26 +1,78 @@
...
+      managerSelectors, 
+      txArgs
+    );
 
     // FEVM: 
     console.log(`registry contract deployed to: ${contract.address}`);
 
+
+    const constructorParams = {
+      parentId: { root: 31415926, route: [] },
```
adlrocha, line 63: We should make this parameterizable, as it was the case before. 
```suggestion
      parentId: { root: chainID, route: [] },
```

scripts/deploy-registry.template
```diff
@@ -1,26 +1,78 @@
...
+      name: ethers.utils.formatBytes32String('Subnet'),
+      ipcGatewayAddr: GATEWAY,
+      consensus: 0,
+      minActivationCollateral: ethers.utils.parseEther("1"),
```
adlrocha, line 67: We may need to set this one to 1FIL not one Ether (I don't think the units match).

src/CloneFactory.sol
```diff
@@ -0,0 +1,19 @@
...
+
+//solhint-disable max-line-length
+//solhint-disable no-inline-assembly
+
+contract CloneFactory {
+    function createClone(address target) internal returns (address result) {
+        bytes20 targetBytes = bytes20(target);
+        assembly {
+            let clone := mload(0x40)
+            mstore(clone, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
```
adlrocha, line 13: Can you explain what are you cloning here and the operation of this contract (I can't of get a sense of what you are trying to do, but without context this code is really hard to understand).

adlrocha (changes requested):
A few comments, specially in the deployment scripts.

scripts/deploy-registry.template
cryptoAtwill: oh yeah, correct, I should remove this

src/CloneFactory.sol
cryptoAtwill: oh oh, this is the standard clone factory assembly code, similar to diamond pattern, I took it from online.

scripts/deploy-registry.template
```diff
@@ -1,26 +1,78 @@
...
+import { ethers } from 'hardhat';
+import { deployContractWithDeployer, getTransactionFees } from './util';
 
-async function main() {
-  try {
+const { getSelectors } = require('./js/diamond.js')
 
-    await hre.run('compile');
+const GATEWAY = '0x4A04DDb64568C5721f6387a6e85FE1585cF7AB9e';
+const LIBMAP = {
```
dnkolegov, line 7: And these are also for debugging?

src/SubnetRegistry.sol
```diff
@@ -27,41 +24,52 @@ contract SubnetRegistry {
...
         }
 
-        subnetAddr = address(new SubnetActor(params));
+        IDiamond.FacetCut[] memory diamondCut = new IDiamond.FacetCut[](2);
 
-        subnets[msg.sender][userNonces[msg.sender]] = subnetAddr;
-        ++userNonces[msg.sender];
+        // set the diamond cut for subnet getter
+        diamondCut[0] = IDiamond.FacetCut({
+            facetAddress: createClone(getterFacet),
```
dnkolegov, line 56: Why do we need this? We do not need that at all if we deploy a new diamond every time. do we?

scripts/deploy-registry.template
cryptoAtwill: they are all 18 decimals right?

src/SubnetRegistry.sol
cryptoAtwill: removed

scripts/deploy-registry.template
cryptoAtwill: removed

scripts/deploy-registry.template
```diff
@@ -1,22 +1,49 @@
...
+      {},
+      txArgs
+    );
+    const getterSelectors = getSelectors(getterFacet);
+    console.log("getter address:", getterFacet.address, "selectors:", getSelectors);
+
+    const managerFacet = await deployContractWithDeployer(
+      deployer,
+      "SubnetActorManagerFacet",
+      LIBMAP,
```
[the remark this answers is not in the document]
cryptoAtwill: oh right, it worked for me because I redefined the LIBMAP. I will check again.

src/SubnetRegistry.sol
```diff
@@ -27,41 +23,52 @@ contract SubnetRegistry {
...
 
-    /// @notice Returns the address of the latest subnet actor
-    /// deployed by a user
-    function latestSubnetDeployed(address owner) external view returns (address subnet) {
-        subnet = subnets[owner][userNonces[owner] - 1];
-        if (subnet == address(0)) {
-            revert ZeroGatewayAddress();
-        }
-    }
+        // we can reduce gas by using clone factory as well, but it
```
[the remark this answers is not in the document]
cryptoAtwill: I think clone factory might actually save some gas, but I left it there as an indicator, but we can remove this as well.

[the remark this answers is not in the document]
dnkolegov: Maybe you should create an issue instead of that line?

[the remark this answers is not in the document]
cryptoAtwill: yeah, I removed that line

dnkolegov (approved):

scripts/deploy-registry.template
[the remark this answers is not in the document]
cryptoAtwill: updated and working!

adlrocha (approved):
LGTM, let's start the integration with the agent if there are no blockers

Issue #71, opened 2023-06-09
Subnet with empty validator set can be activated


dnkolegov:
Let's say `minActivationCollateral` = 10, validator 1 stakes 3, validator 2 stakes 7.

After validator 1 calling `join`: `totalStake` = 3, `validatorSet` = {}
After validator 2 calling `join`: `totalStake` = 10, `validatorSet` = {} but since `totalStake >= minActivationCollateral` then `addStake` is called and changes the status of the subnet to `Active`. Then subnet actor also changes the subnet status to `Active`.

As a result, the subnet with an empty validator set is active.

dnkolegov:
At present, "Active" means that a subnet has enough collateral. In the current design, even a subnet without validators can be active. This is also implemented in Gateway's `register` function that activates the Subnet. Also, the Gateway actor knows nothing about the number of validators in a subnet.
If we want explicitly signal to Gateway when a subnet should be activated/deactivated, we should use calls like `activateSubnet` and `deactivateSubnet` of the Gateway actor.

Closing for now.

dnkolegov:
Another problem with the current design is that the subnet collateral can fall behind the minimum and cause the subnet to be deactivated. However, if every single validator has to put up at least as much as the minimum required collateral, then falling behind the threshold means all validators have abandoned the subnet.

https://github.com/consensus-shipyard/ipc-solidity-actors/pull/201#discussion_r1325783577

aakoshh:
> After validator 1 calling join: totalStake = 3, validatorSet = {}

That sounds like a bug, shouldn't `join` add the guy to the `validatorSet`?

>  In the current design, even a subnet without validators can be active. 

That sounds strange, who put the collateral down if there are no validators? How does this design make any sense?

dnkolegov:
> According to this implementation, it should not since the validator's stake is less than `minActivationCollateral`.

I agree that this is a bug the problem is that we do not have a spec.

adlrocha:
I have an alternative view of this issue (I am not claiming it is the right one, but is how I was thinking about it when I first implemented this). 
- The gateway is the gatekeeper of IPC. It is responsible for keeping the collateral for the subnet and enforcing a set of minimum requirements for all its child subnets. The rules are simple right now: if a child subnet has enough collateral according to the threshold set for this IPC parent, the subnet is `Active` and can interact with IPC, if not, it is in `Inactive`. 
[...]

PR #156, opened 2023-07-25
Fix issue 71

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/71

### DIFF
--- a/src/gateway/GatewayManagerFacet.sol
+++ b/src/gateway/GatewayManagerFacet.sol
@@ -68,6 +68,15 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         }
 
         subnet.stake += msg.value;
+    }
+
+    /// @notice activateSubnet - activates the subnet if it has sufficient collateral
+    function activateSubnet() external {
+        (bool registered, Subnet storage subnet) = LibGateway.getSubnet(msg.sender);
+
+        if (!registered) {
+            revert NotRegisteredSubnet();
+        }
 
         if (subnet.status == Status.Inactive) {
             if (subnet.stake >= s.minStake) {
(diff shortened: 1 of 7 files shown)


src/lib/LibSubnetActorStorage.sol
```diff
@@ -66,6 +64,7 @@ struct SubnetActorStorage {
     bytes32 currentSubnetHash;
     /// @notice Minimal number of validators required for the subnet to be able to validate new blocks.
     uint64 minValidators;
+    /// @notice Minimal number that a validator requires to stake to be added to the validator set of the subnet.
```
adlrocha, line 67:
```suggestion
    /// @notice Minimal collateral that a validator requires to stake to be added to the validator set of the subnet.
```

src/gateway/GatewayManagerFacet.sol
```diff
@@ -68,6 +68,15 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         }
 
         subnet.stake += msg.value;
+    }
+
+    /// @notice activateSubnet - activates the subnet if it has sufficient collateral
```
adlrocha, line 73: Isn't this redundant with the function above or am I missing something?

test/SubnetActorDiamond.t.sol
```diff
@@ -695,7 +695,7 @@ contract SubnetActorDiamondTest is Test {
     }
 
     function callback() public view {
-        console.log("callback called");
+        // console.log("callback called");
```
adlrocha, line 698: Is this on purpose?

adlrocha (approved):
I am having a hard time understanding how this fix solves the issue described in #71. Do you mind describing the solution high-level? (we can sync if you think it may be easier)

test/SubnetActorDiamond.t.sol
dnkolegov: image

dnkolegov: Just to disable this

PR #160, opened 2023-07-26
Fix bottom up checkpoint period

cryptoAtwill:
Bottom up checkpoint period is not set when subnet was created. This PR fixes that.


adlrocha (approved):
Good catch, thanks! Be sure to fix CI before merging

dnkolegov (approved):

PR #161, opened 2023-07-27
Fix list bottom up

cryptoAtwill:
Using multiples of bottom up checkpoint period to list bottom up checkpoints.

### DIFF
--- a/src/subnet/SubnetActorGetterFacet.sol
+++ b/src/subnet/SubnetActorGetterFacet.sol
@@ -164,10 +164,13 @@ contract SubnetActorGetterFacet {
     ) external view returns (BottomUpCheckpoint[] memory) {
         uint64 period = s.bottomUpCheckPeriod;
 
-        uint64 size = (toEpoch - fromEpoch) / period;
+        uint64 from = (fromEpoch / period) * period;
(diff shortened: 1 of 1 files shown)


src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -162,13 +162,17 @@ contract SubnetActorGetterFacet {
         uint64 fromEpoch,
         uint64 toEpoch
     ) external view returns (BottomUpCheckpoint[] memory) {
-        uint64 size = toEpoch - fromEpoch;
+        uint64 period = s.bottomUpCheckPeriod;
```
dnkolegov, line 165: Do we not need to validate fromEpoch and toEpoch?

cryptoAtwill: I assume the caller should be responsible for providing valid epochs.

adlrocha (approved):
Fixed CI. We should include a new test every time we include a new getter as a sanity-check. We need to have high coverage before the audit. I will get this one merged so we have the diamond architecture in main, but let's consider it for a follow-up. Thanks!

PR #163, opened 2023-07-28
Minor fixes

dnkolegov:
This PR:

- disable printing
- remove unused functions


test/GatewayDiamond.t.sol
```diff
@@ -2333,7 +2333,7 @@ contract GatewayDiamondDeploymentTest is StdInvariant, Test {
     }
 
     function callback() public view {
-        console.log("callback called");
+        // console.log("callback called");
```
adlrocha, line 2336: Is this on purpose? Why do we need to comment it (and not even remove it all the way?)

adlrocha (approved):

PR #171, opened 2023-08-02
Add messenger facet

dnkolegov:
This PR adds a messenger facet.


aakoshh:
Why do you check in `.storage-layouts` but not the full JSON which includes the ABI?

scripts/deploy-registry.template
```diff
@@ -27,7 +27,6 @@ async function main() {
       deployer,
       "SubnetActorManagerFacet",
       {
-        AccountHelper: LIBMAP['AccountHelper'],
```
adlrocha, line 30: Why was this removed?

adlrocha (approved):
LGTM. If you tested the deployment and it works feel free to merge. Let me know if you need help testing the deployment.

scripts/deploy-registry.template
dnkolegov: Because we do not need it after removing `isAccount` check and the compiler does not compile the code if it sees unnecessary libraries.

dnkolegov:
> Why do you check in `.storage-layouts` but not the full JSON, which includes the ABI?

We use a ready module for this, and it works with files like those. You can think about it like `Cargo.lock` but for the storage state.

cryptoAtwill (approved):

PR #181, opened 2023-08-07
Add setValidatorNetAddr

dnkolegov:
This PR adds `setValidatorNetAddr` and `setValidatorWorkerAddr` method.


adlrocha (approved):
Initially I thought about having a single method to edit a validators info all at once, but I think for now having them separately makes sense. Thanks!

We will need to integrate this one to the agent, I think this could be a good first issue for you get on-boarded in the agent: https://github.com/consensus-shipyard/ipc-agent/issues/286

dnkolegov:
> consensus-shipyard/ipc-agent#286

I think they should be separate. It is more gas efficient. Separation affects the size of the contract, but we have Diamond infrastructure for that.

PR #182, opened 2023-08-07
Add pagination

dnkolegov:
This PR adds a pagination method for validators.

### DIFF
--- a/src/subnet/SubnetActorGetterFacet.sol
+++ b/src/subnet/SubnetActorGetterFacet.sol
@@ -109,7 +109,7 @@ contract SubnetActorGetterFacet {
     }
 
     /// @notice get all the validators in the subnet.
-    function getAllValidators() external view returns (address[] memory) {
+    function getValidators() external view returns (address[] memory) {
         uint256 length = s.validators.length();
         address[] memory result = new address[](length);
 
@@ -123,8 +123,11 @@ contract SubnetActorGetterFacet {
         return result;
     }
 
-    /// @notice get n validators of the subnet from a range offset to limit.
-    function getValidators(uint256 offset, uint256 limit) external view returns (address[] memory, uint256) {
+    /// @notice get no more than `limit` number of validators starting from the validator with index `offset`
+    /// @dev It returns an empty array[] and 0 if there are no validators to return according to the input parameters
+    /// @return the array of validators, the size of that array is no more than `limit`
+    /// @return the next `offset`
+    function getRangeOfValidators(uint256 offset, uint256 limit) external view returns (address[] memory, uint256) {
         uint256 n = s.validators.length();
         address[] memory empty = new address[](0);
         if (limit == 0) {
--- a/src/subnet/SubnetActorGetterFacet.sol
+++ b/src/subnet/SubnetActorGetterFacet.sol
@@ -125,6 +125,8 @@ contract SubnetActorGetterFacet {
 
     /// @notice get no more than `limit` number of validators starting from the validator with index `offset`
     /// @dev It returns an empty array[] and 0 if there are no validators to return according to the input parameters
+    /// @param offset The first index of the first validator to return
+    /// @param limit The maximum number of validators to return
     /// @return the array of validators, the size of that array is no more than `limit`
     /// @return the next `offset`
     function getRangeOfValidators(uint256 offset, uint256 limit) external view returns (address[] memory, uint256) {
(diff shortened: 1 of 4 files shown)


src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -123,6 +123,32 @@ contract SubnetActorGetterFacet {
         return result;
     }
 
+    /// @notice get n validators in the subnet starting with index.
+    function getValidators(uint256 offset, uint256 limit) external view returns (address[] memory, uint256) {
```
adlrocha, line 127: Can we rename this function instead of calling it `getValidators`? Using `getValidators` for this one would require changing the agent integration. 
```suggestion
    function getRangeOfValidators(uint256 offset, uint256 limit) external view returns (address[] memory, uint256) {
```

src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -108,8 +109,7 @@ contract SubnetActorGetterFacet {
     }
 
     /// @notice get all the validators in the subnet.
-    /// TODO: we can introduce pagination
-    function getValidators() external view returns (address[] memory) {
+    function getAllValidators() external view returns (address[] memory) {
```
adlrocha, line 112: Let's keep the name to avoid having to change the agent.
```suggestion
    function getValidators() external view returns (address[] memory) {
```

src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -123,6 +123,32 @@ contract SubnetActorGetterFacet {
...
+        if (n <= offset) {
+            return (empty, 0);
+        }
+
+        if (limit > n - offset) {
+            limit = n - offset;
+        }
+        address[] memory result = new address[](limit);
+
+        for (uint256 i = 0; i < limit; ) {
```
adlrocha, line 142: Wouldn't it be easier to check the limit above and instead of accessing index `(i+offset)`, iterate through the real index directly? 
```suggestion
        for (uint256 i = offset; i < limit; ) {
```

No strong opinions, it just seems more straightforward.

src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -123,6 +123,32 @@ contract SubnetActorGetterFacet {
...
+            result[i] = s.validators.at(i + offset);
+            unchecked {
+                ++i;
+            }
+        }
+
+        return (result, offset + limit);
```
adlrocha, line 149: Why do you return the `offset + limit`? Let's maybe add a comment to explain the return values.

test/SubnetActorDiamond.t.sol
```diff
@@ -292,12 +292,86 @@ contract SubnetActorDiamondTest is Test {
...
+        require(saGetter.validatorCount() == 7);
+        require(saGetter.getAllValidators().length == 7);
+        require(saGetter.getValidatorSet().validators.length == 7);
+
+        address[] memory result;
+        uint256 offset;
+
+        (result, offset) = saGetter.getValidators(0, 2);
+        require(result.length == 2);
+        require(offset == 2);
```
adlrocha, line 327: Is this really the offset or the final index for the validator returned? 
```suggestion
        require(index == 2);
```

src/subnet/SubnetActorGetterFacet.sol
dnkolegov: Fixed

test/SubnetActorDiamond.t.sol
dnkolegov: That is next offset - the index of the next validator in the next call

src/subnet/SubnetActorGetterFacet.sol
adlrocha: It makes sense, let's just document it so other devs know how to use it.

src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -123,6 +123,37 @@ contract SubnetActorGetterFacet {
         return result;
     }
 
+    /// @notice get no more than `limit` number of validators starting from the validator with index `offset`
+    /// @dev It returns an empty array[] and 0 if there are no validators to return according to the input parameters
+    /// @param offset The first index of the first validator to return
+    /// @param limit The maximum number of validators to return
+    /// @return the array of validators, the size of that array is no more than `limit`
+    /// @return the next `offset`
```
adlrocha, line 131:
```suggestion
    /// @return the next `offset` that needs to query the next range of validators
```

adlrocha (approved):
Thank you for addressing the comments. Now LGTM.

PR #184, opened 2023-08-08
get chain-id automatically in deploy-ipc

adlrocha:
This PR updates the deployment scripts in a way that by default or if `NETWORK=auto` is enabled, the deployment scripts automatically fetch the chain ID of the network `RPC_URL` is pointing to in order to deploy the IPC contracts with the right ID.

### DIFF
--- a/ops/chain-id.sh
+++ b/ops/chain-id.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# Check if RPC_URL is set
+if [[ -z "$RPC_URL" ]]; then
+    echo "RPC_URL is not set. Sourcing .env file..."
+    source .env
+fi
+
+# Make a JSON-RPC call to get the chain ID using curl
+response=$(curl -s -X POST -H "Content-Type: application/json" --data '{
+  "jsonrpc":"2.0",
+  "method":"eth_chainId",
+  "params":[],
+  "id":1
+}' $RPC_URL)
+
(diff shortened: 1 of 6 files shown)


ops/chain-id.sh
```diff
@@ -0,0 +1,24 @@
...
+# Make a JSON-RPC call to get the chain ID using curl
+response=$(curl -s -X POST -H "Content-Type: application/json" --data '{
+  "jsonrpc":"2.0",
+  "method":"eth_chainId",
+  "params":[],
+  "id":1
+}' $RPC_URL)
+
+# Extract the chain ID from the response using jq (ensure jq is installed)
+chain_id=$(echo $response | jq -r '.result')
```
dnkolegov, line 18: If it is possible I would suggest checking that response is valid and `chain_id` contains only digits. And print an error otherwise.

dnkolegov (changes requested):

ops/chain-id.sh
adlrocha: I already check if it is valid in the hardhat config and in the deploy-gateway template, but I added an additional sanity-check.

dnkolegov: Thank you. I was worrying about the injection of malicious environment variables leading to a malicious RPC that could return executable code or something like that.

dnkolegov (approved):

PR #186, opened 2023-08-09
Add examples of using SendCrossMessage in contract

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/135 by introducing a simple contract that triggers a cross-net message to some subnet when the target function is called.

### DIFF
--- a/examples/contracts/Messenger.sol
+++ b/examples/contracts/Messenger.sol
@@ -62,13 +62,7 @@ contract Messenger {
         });
 
         // emit event and increase nonce
-        emit MessageSent({
-            from: from,
-            to:to,
-            nonce: nonce++,
-            messageBody:messageBody
-        });
-
+        emit MessageSent({from: from, to: to, nonce: nonce++, messageBody: messageBody});
 
         // slither-disable-next-line arbitrary-send-eth
         messenger.sendCrossMessage{value: DEFAULT_CROSS_MSG_FEE}(crossMsg);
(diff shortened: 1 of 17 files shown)


examples/contracts/Messenger.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    constructor(address _gateway){
+        if (_gateway == address(0)) {
+            revert GatewayCannotBeZero();
+        }
+        sender = new GatewayMessengerFacet();
+        sender = GatewayMessengerFacet(address(_gateway));
+    }
+
+    function sendMessage(CrossMsg calldata crossMsg) external payable {
```
adlrocha, line 22: We can keep this contract as an example, but as discussed sync, the idea is to have a sample contract like the `Storage.sol` of Remix that triggers a specific cross-net message to illustrate how this can be used from a contract. This will also help us test the integration end-to-end. 

We can consider the following scenario: 
- We deploy a Storage.sol in one subnet. Then we have our sample contract that is a CrossStorage.sol point to the first one that whenever someone changes the value stored, it triggers a cross-net message to store the same value in the `Storage.sol`. 
- We can consider any kind of ownership or access policy we want. Instead of having a single storage slot, we could have a map that stores the value for the specific owner, and only allow if it comes from a cross-net message.

The idea is to start exploring how the on-chain primitives of IPC can be used and showcase them.

adlrocha (commented):
As mentioned in the comments, the goal of this is to help us with the end-to-end testing of this function, and to start exploring how the on-chain primitives of IPC can be used and showcase them.

- Let's try to think a more complex example of a contract that leverages cross-net primitives. 
- Let's add a README in the examples software with a description of the contracts, what they do, how IPC primitives are used, etc.

PR #187, opened 2023-08-10
Log event

cryptoAtwill:
# Motivation
Currently, the manager is quite difficult to debug as it does not expose intermediate states nor exposes events. Whenever the voting goes wrong, it is almost impossible to debug where goes wrong. This PR attempts to expose internal events so that we can track what is going on.

# Solution
Adding `BottomUpCheckpointSubmitted` and `BottomUpCheckpointExecuted` events to expose the internal process.

### DIFF
--- a/src/subnet/SubnetActorManagerFacet.sol
+++ b/src/subnet/SubnetActorManagerFacet.sol
@@ -30,6 +30,16 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     using EpochVoteSubmissionHelper for EpochVoteSubmission;
     using FvmAddressHelper for FvmAddress;
 
+    event BottomUpCheckpointSubmitted(
+        BottomUpCheckpoint checkpoint,
+        address submitter,
+        bool shouldExecute
+    );
+    event BottomUpCheckpointExecuted(
+        uint64 epoch,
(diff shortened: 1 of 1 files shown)


dnkolegov (approved):

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -128,8 +132,11 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
             submitterWeight: s.stake[msg.sender]
         });
 
+        emit BottomUpCheckpointSubmitted(checkpoint, msg.sender);
```
dnkolegov, line 135: Slither complains about possible reentrancies: try to emit first, then change the state

dnkolegov (changes requested):

adlrocha (commented):
This makes total sense! Thank you @cryptoAtwill. Once you address @dnkolegov comment and he approves, from my side feel free to merge.

src/subnet/SubnetActorManagerFacet.sol
cryptoAtwill: This is resolved, please help take another look.

dnkolegov (approved):

PR #190, opened 2023-08-23
Arrange top down messages by height

cryptoAtwill:
To integrate with Fendermint, we are arranging the top down checkpoint by height. Key changes:
- Changing `topDownMsgs` from array to `mapping(uint256 =>array)` with block height as key.
- Update all the getter functions, removed unused ones.

### DIFF
--- a/src/gateway/GatewayGetterFacet.sol
+++ b/src/gateway/GatewayGetterFacet.sol
@@ -3,7 +3,7 @@ pragma solidity 0.8.19;
 
 import {CrossMsg, BottomUpCheckpoint, StorableMsg} from "../structs/Checkpoint.sol";
 import {EpochVoteTopDownSubmission} from "../structs/EpochVoteSubmission.sol";
-import {SubnetID, Subnet} from "../structs/Subnet.sol";
+import {SubnetID, Subnet, SubnetDTO} from "../structs/Subnet.sol";
 import {CheckpointHelper} from "../lib/CheckpointHelper.sol";
 import {LibGateway} from "../lib/LibGateway.sol";
 import {GatewayActorStorage} from "../lib/LibGatewayActorStorage.sol";
@@ -57,7 +57,7 @@ contract GatewayGetterFacet {
     /// @param subnetId the id of the subnet
     /// @return found whether the subnet exists
     /// @return subnet -  the subnet struct
-    function getSubnet(SubnetID calldata subnetId) external view returns (bool, Subnet memory) {
+    function getSubnet(SubnetID calldata subnetId) external view returns (bool, SubnetDTO memory) {
         // slither-disable-next-line unused-return
         return LibGateway.getSubnet(subnetId);
     }
--- a/src/gateway/GatewayGetterFacet.sol
+++ b/src/gateway/GatewayGetterFacet.sol
@@ -70,37 +70,50 @@ contract GatewayGetterFacet {
     function getSubnetTopDownMsgsLength(SubnetID memory subnetId) external view returns (uint256) {
         // slither-disable-next-line unused-return
         (, Subnet storage subnet) = LibGateway.getSubnet(subnetId);
-        return subnet.topDownMsgs.length;
+        return subnet.topDownNonce;
     }
 
-    /// @notice get the top-down message at the given index for the given subnet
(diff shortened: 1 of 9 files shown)


src/gateway/GatewayGetterFacet.sol
```diff
@@ -70,41 +70,19 @@ contract GatewayGetterFacet {
     function getSubnetTopDownMsgsLength(SubnetID memory subnetId) external view returns (uint256) {
         // slither-disable-next-line unused-return
         (, Subnet storage subnet) = LibGateway.getSubnet(subnetId);
-        return subnet.topDownMsgs.length;
+        return subnet.topDownNonce;
```
dnkolegov, line 73: Why `getSubnetTopDownMsgsLength` returns `topDownNonce`?

src/gateway/GatewayGetterFacet.sol
```diff
@@ -70,41 +70,19 @@ contract GatewayGetterFacet {
...
 
-    /// @notice get the top-down message at the given index for the given subnet
-    function getSubnetTopDownMsg(SubnetID memory subnetId, uint256 index) external view returns (CrossMsg memory) {
-        // slither-disable-next-line unused-return
-        (, Subnet storage subnet) = LibGateway.getSubnet(subnetId);
-        return subnet.topDownMsgs[index];
-    }
-
-    /// @notice get the list of top down messages from nonce, we may also consider introducing pagination.
+    /// @notice get the list of top down messages from block number, we may also consider introducing pagination.
```
dnkolegov, line 76: Why is getSubnetTopDownMsg removed?

src/gateway/GatewayGetterFacet.sol
cryptoAtwill: With every new message, the nonce is added by one, the current nonce should be equal to the top down message length.

dnkolegov: Could you add this comment to the code please?

cryptoAtwill: Updated

src/lib/LibGateway.sol
```diff
@@ -77,6 +78,51 @@ library LibGateway {
...
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+
+        // invalid from block number
+        if (fromBlock > toBlock) {
+            return new CrossMsg[](0);
+        }
+
+        bytes32 subnetHash = subnetId.toHash();
+        uint256 msgLength = 0;
+        for (uint256 i = fromBlock; i <= toBlock; ) {
```
adlrocha, line 99: Dumb question, I guess there is no work around this, right? We necessarily need to make this first iteration to compute the length of the array.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -12,6 +12,9 @@ struct GatewayActorStorage {
     /// @notice List of subnets
     /// SubnetID => Subnet
     mapping(bytes32 => Subnet) subnets;
+    /// @notice a mapping of block number to cross messages
+    /// SubnetID => blockNumber => messages
+    mapping(bytes32 => mapping(uint256 => CrossMsg[])) topDownMsgs;
```
adlrocha, line 17: Out of a curiosity, why did you do it in a nested map instead of adding a map inside a subnet? I guess this is fine, I am just worried about having outdated state in this mapping when a subnet is destroyed.

adlrocha (approved):
LGTM, but please have a look at my comment (sanity-check before merging). Thanks!

src/lib/LibGateway.sol
cryptoAtwill: yeah, the first iteration is to get the total length.

src/lib/LibGatewayActorStorage.sol
cryptoAtwill: It's because if I put this in `Subnet` struct, it won't compile. Many functions are returning `memory Subnet` and nested mapping in `Subnet` cannot be converted to memory.

adlrocha:
I guess the tests being red is the issue @dnkolegov and you identified, right? I leave you both click the merge button when ready :)

PR #192, opened 2023-08-24
Multi-signature ECDSA

dnkolegov:
This PR implements multisignature ECDSA.


src/lib/LibMultisignatureChecker.sol
```diff
@@ -0,0 +1,126 @@
...
+        bytes32 r;
+        bytes32 s;
+
+        for (uint256 i = 0; i < signaturesNumber; ) {
+            (v, r, s) = parseSignature(signatures, i);
+            (address recovered, ECDSA.RecoverError ecdsaErr, ) = ECDSA.tryRecover({hash: hash, v: v, r: r, s: s});
+            if (ecdsaErr != ECDSA.RecoverError.NoError) {
+                return (!valid, Error.InvalidSignature);
+            }
+            if (recovered != signatories[i]) {
```
aakoshh, line 113: Where do the addresses get lined up with the signatures? I can [only see](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/218/files#diff-c33605fbd68a5aeb1782c3068dcb37d917cbc29d47cc2b38c3a06fd0543bb25eR237) signatures returned.

aakoshh (commented):
This looks cool, I only have two questions:
1. Who is supposed to line up signatures with addresses? Does it need a modification of the checkpoint?
2. Is this going to be followed up by another PR that implements more of https://github.com/consensus-shipyard/fendermint/issues/192 ?

src/lib/LibMultisignatureChecker.sol
```diff
@@ -0,0 +1,126 @@
...
```
adlrocha, line 71: This assumes that the signatories need to be passed in order for the multisig verification to work, right? Maybe we should be explicit about this in the function description?

src/lib/LibMultisignatureChecker.sol
adlrocha: This links to my comment above, we should probably return a bundle of multisig + ordered addresses to avoid potential blunders.

tools/check_coverage.sh
```diff
@@ -6,7 +6,7 @@ measurements[0]="lines"
 measurements[1]="functions"
 measurements[2]="branches"
 
-threshold=45
+threshold=30
```
adlrocha, line 9: I am afraid we will keep cheating with the threshold until we reach a decent coverage :see_no_evil:

src/lib/LibMultisignatureChecker.sol
```diff
@@ -0,0 +1,126 @@
...
+    function countSignatures(bytes memory signatures) public pure returns (uint256) {
+        uint256 len = signatures.length;
+        return (len % SIGNATURE_LENGTH) == 0 ? (len / SIGNATURE_LENGTH) : 0;
+    }
+
+    /**
+     * @dev Checks if a multi-signature is valid for a given message hash, set of signatories, and set of signatures.
+     * Signatures are validated using `ECDSA.recover`.
+     */
+    function isValidMultiSignature(
```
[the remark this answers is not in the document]
adlrocha: +1, this function seems redundant.

adlrocha:
> Is this going to be followed up by another PR that implements more of https://github.com/consensus-shipyard/fendermint/issues/192 ?

I probably can reply to this one. This is what we are planning next :)

src/lib/LibMultisignatureChecker.sol
dnkolegov: Fixed

src/lib/LibMultisignatureChecker.sol
dnkolegov: Fixed

adlrocha (approved):
Thank you for the fixes. LGTM

src/lib/LibMultisignatureChecker.sol
```diff
@@ -0,0 +1,94 @@
...
+     * @param threshold The number that must be reach to consider `signatures` valid.
+     * @param hash of the verified data.
+     * @param signatures Packed signatures. Each signature is in `({bytes32 r}{bytes32 s}{uint8 v})` format.
+     */
+    function isValidWeightedMultiSignature(
+        address[] memory signatories,
+        uint256[] memory weights,
+        uint256 threshold,
+        bytes32 hash,
+        bytes memory signatures
```
aakoshh, line 58: Just curious: why do you want the concatenation of signatures? Is this some generalisation of a multi-sig that could have other formats, like bitfields?

test/LibMultisignatureChecker.t.sol
```diff
@@ -0,0 +1,239 @@
...
+
+        for (uint256 i = 0; i < 2; i++) {
+            (uint8 v, bytes32 r, bytes32 s) = vm.sign(PRIVATE_KEY_BASE + i, hash);
+            bytes memory signature = abi.encodePacked(r, s, v);
+            multisignatureBytes = bytes.concat(multisignatureBytes, signature);
+            weights[i] = 10;
+        }
+
+        signers[0] = vm.addr(PRIVATE_KEY_BASE + 1);
+        signers[1] = vm.addr(PRIVATE_KEY_BASE);
```
aakoshh, line 169: Why isn't this the other way around? The loop goes from 0 to 1 and signs with `PRIVATE_KEY_BASE + i`, so why isn't `signers[0] = vm.addr(PRIVATE_KEY_BASE)`?

aakoshh: Oh, I see because you are testing the invalid case. Would be worth leaving a comment that this is the crucial bit that screws it up.

aakoshh (approved):
LGTM!

src/lib/LibMultisignatureChecker.sol
dnkolegov: Because there is no such type to work easily with 65-byte signatures. Ideally, it should be `bytes65[] signatures`.

aakoshh: True, but the "bundle" method (forgot where it was) returns `[]bytes` IIRC, so we can clearly work with arrays, but have to go the extra mile of concatenating before passing it back to this. 

You can always validate that members in the input array are 65 bytes long.

dnkolegov: Ok. I will fix that.

dnkolegov: Thank you Akosh.

Fixed here - https://github.com/consensus-shipyard/ipc-solidity-actors/pull/219/commits/667bae3f77746a4c84a0861e978d8db2ced52b63

I like this new one.

cc @adlrocha

test/LibMultisignatureChecker.t.sol
dnkolegov: Fixed

PR #193, opened 2023-08-29
Commit parent finality

cryptoAtwill:
Initial attempt to store and retrieve IPC parent finality in gateway contract. The updates are as follows:
- Move `setMembership` to `GatewayLib` and remove the corresponding external function call in `GatewayManagerFacet`.
- Add `commitParentFinality` to `RouterFacet`.
- Add `queryParentFinality` to `GatewayGetterFacet`.

In follow up PR will clean up the code by removing submit top down and bottom up checkpoint and all their internal functions.


src/gateway/GatewayRouterFacet.sol
```diff
@@ -142,6 +142,20 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         _applyMessages(SubnetID(0, new address[](0)), topDownMsgs);
     }
 
+    /// @notice commit the ipc parent finality into storage
+    function commitParentFinality(
```
dnkolegov, line 146: Are we going to have `commitChildFinality` ? Why just not `commitFinality`?

src/gateway/GatewayGetterFacet.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity 0.8.19;
 
-import {CrossMsg, BottomUpCheckpoint, StorableMsg} from "../structs/Checkpoint.sol";
+import {CrossMsg, BottomUpCheckpoint, StorableMsg, ParentFinality} from "../structs/Checkpoint.sol";
```
dnkolegov, line 4: Do we have other types of Finality besides parent?

cryptoAtwill: Yep, there might be for bottom up or others in the future.

test/GatewayDiamond.t.sol
```diff
@@ -1633,55 +1633,58 @@ contract GatewayDiamondDeploymentTest is StdInvariant, Test {
...
-        gwManager.setMembership(validators, weights);
+        gwRouter.commitParentFinality(finality, new CrossMsg[](0), validators, weights);
     }
 
-    function testGatewayDiamond_SetMembership_Fails_ZeroWeight() public {
-        address[] memory validators = new address[](1);
-        validators[0] = vm.addr(100);
-        uint256[] memory weights = new uint256[](1);
-        weights[0] = 0;
+    // function testGatewayDiamond_SetMembership_Fails_ValidatorsAndWeightsNotEqual() public {
```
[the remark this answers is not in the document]
cryptoAtwill: I have updated them for the commit finality, in the future, they should be removed

test/GatewayDiamond.t.sol
```diff
@@ -1967,46 +1970,46 @@ contract GatewayDiamondDeploymentTest is StdInvariant, Test {
...
-
-        TopDownCheckpoint memory checkpoint = TopDownCheckpoint({
-            epoch: DEFAULT_CHECKPOINT_PERIOD,
-            topDownMsgs: topDownMsgs
-        });
-
-        vm.prank(validators[0]);
-        gwRouter.submitTopDownCheckpoint(checkpoint);
-    }
+    // function testGatewayDiamond_SubmitTopDownCheckpoint_BigNumberOfMessages() public {
```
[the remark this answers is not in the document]
cryptoAtwill: I have updated them for the commit finality, in the future, they should be removed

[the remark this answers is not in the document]
dnkolegov: Could you please add this comment right into the code (in TODO) to not forget about that in the future?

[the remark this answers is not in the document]
adlrocha: We should remove this right away. We are not going to support `submitTopDownCheckpoint` anymore, so if we want to check this code we can always look at an outdated commit.

src/lib/LibGateway.sol
```diff
@@ -35,6 +35,59 @@ library LibGateway {
...
     }
 
+    /// @notice obtain the ipc parent finality at certain block number
+    /// @param blockNumber - the block number to obtain the finality
+    function getParentFinality(uint256 blockNumber) internal view returns (ParentFinality memory) {
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+        return s.parentFinalities[blockNumber];
+    }
+
+    /// @notice commit the ipc parent finality into storage
```
adlrocha, line 45:
```suggestion
    /// @notice commit the ipc parent finality into storage.
    /// Committing the finality triggers the execution of top-down cross-net
    /// messages and updates the membership info.
```

src/lib/LibGateway.sol
```diff
@@ -35,6 +35,59 @@ library LibGateway {
...
+        for (uint256 validatorIndex = 0; validatorIndex < validatorsLength; ) {
+            address validatorAddress = validators[validatorIndex];
+            if (validatorAddress != address(0)) {
+                uint256 validatorWeight = weights[validatorIndex];
+
+                if (validatorWeight == 0) {
+                    revert ValidatorWeightIsZero();
+                }
+
+                s.validatorSet[s.validatorNonce][validatorAddress] = validatorWeight;
```
adlrocha, line 78: I don't think we need to address it immediately, but we will need to garbage collect the membership info instead of keeping all the history. We may be able to garbage collect outdated info.

adlrocha (changes requested):
LGTM. The reason I request the change is because to wrap it up we should remove https://github.com/consensus-shipyard/ipc-solidity-actors/blob/cc109a57d62755cf849028a366258c38c04e49c8/src/gateway/GatewayManagerFacet.sol#L25. I don't think is needed anymore. If we need the `topDownGenesisEpoch` for something, we can set it directly when the first proof of finality is committed, but it can be kind of a mess having to also call this `initGenesisEpoch` implicitly when we are actually doing it implicitly when committing finality.

We can discuss sync if it is not clear.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -15,6 +15,8 @@ struct GatewayActorStorage {
     /// @notice a mapping of block number to cross messages
     /// SubnetID => blockNumber => messages
     mapping(bytes32 => mapping(uint256 => CrossMsg[])) topDownMsgs;
+    /// @notice The parent finalities. Key is the block number, value is the finality struct.
+    mapping(uint256 => ParentFinality) parentFinalities;
```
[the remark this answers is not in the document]
dnkolegov: I asked about naming and @cryptoAtwill told me we have other finalities in the future like `childFinality`. If this is still true then we may need to use `parent`/`child`/`smth` prefix.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -142,6 +142,20 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        ParentFinality calldata finality,
+        CrossMsg[] calldata crossMsgs,
+        address[] calldata validators,
+        uint256[] calldata weights
+    ) external systemActorOnly {
+        LibGateway.commitParentFinality(finality);
```
adlrocha, line 152: Should we do some sanity-checks here as the ones we have in Fendermint? (like that the height is over the previous height or something like that?)

adlrocha:
@cryptoAtwill, we should also remove all the code around `TopDownVoting`, `_submitTopDownVote`, etc. This is no longer necessary. I'll check the changes required to apply_messages and we can discuss if to apply them here or in some other PR.

cryptoAtwill:
@adlrocha sounds good, I will do all those deletion related work in a PR on top of this so that the changes are not mixed in this PR.

test/GatewayDiamond.t.sol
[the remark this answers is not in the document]
cryptoAtwill: I removed already

cryptoAtwill:
@adlrocha I will create a separate PR to delete other methods, such as `initGenesis`. Because keep on deleting them will fail the unit tests and the PR get bigger. Another PR might be neater.

src/gateway/GatewayRouterFacet.sol
cryptoAtwill: yep, they are added

src/lib/LibGateway.sol
cryptoAtwill: I added another variable to track the block height of the last committed height, we can commit and delete at the same time. I can implement this logic in follow up PRs.

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -26,7 +26,7 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
      */
     function sendCrossMessage(CrossMsg calldata crossMsg) external payable hasFee {
         if (crossMsg.message.value != msg.value - s.crossMsgFee) {
-            revert NotEnoughFunds();
+            revert InvalidCrossMsgValue();
```
dnkolegov, line 29: @adlrocha what do you think? I think the previous error `NotEnoughFunds()` was correct

cryptoAtwill: this is supposed to equal right?, if `msg.value` is really small, then `crossMsg.message.value` is more than required. Even though `hasFee` performs the check, but not enough fund is only valid with the `hasFee` is present. But just based on the code, it's requiring exact equal.

dnkolegov (approved):

src/gateway/GatewayRouterFacet.sol
```diff
@@ -18,171 +18,31 @@ import {LibVoting} from "../lib/LibVoting.sol";
...
     using FilAddress for address;
     using SubnetIDHelper for SubnetID;
     using CrossMsgHelper for CrossMsg;
-    using CheckpointHelper for BottomUpCheckpoint;
-    using CheckpointHelper for TopDownCheckpoint;
+    using FvmAddressHelper for FvmAddress;
     using StorableMsgHelper for StorableMsg;
 
-    /// @notice submit a checkpoint in the gateway. Called from a subnet once the checkpoint is voted for and reaches majority
-    function commitChildCheck(BottomUpCheckpoint calldata commit) external {
```
adlrocha, line 32: We should also keep this one in the gateway.

adlrocha (changes requested):
@cryptoAtwill, few more changes before we merge, I can address them myself if you want. Let me know.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -18,171 +18,31 @@ import {LibVoting} from "../lib/LibVoting.sol";
...
-
-        if (subnet.circSupply < totalValue) {
-            revert NotEnoughSubnetCircSupply();
-        }
-
-        subnet.circSupply -= totalValue;
-
-        subnet.prevCheckpoint = commit;
+    /// @notice commit the ipc parent finality into storage
+    function commitParentFinality(
```
dnkolegov, line 33: Why is this method not in the GatewayActor interface?

src/interfaces/ISubnetActor.sol
```diff
@@ -18,13 +18,6 @@ interface ISubnetActor {
     /// Unregister the subnet from the hierarchy, making it no longer discoverable.
     function kill() external;
 
-    /// SubmitCheckpoint accepts signed checkpoint votes for miners.
-    ///
-    /// This functions verifies that the checkpoint is valid before
-    /// propagating it for commitment to the IPC gateway. It expects at least
-    /// votes from 2/3 of miners with collateral.
-    function submitCheckpoint(BottomUpCheckpoint calldata checkpoint) external;
```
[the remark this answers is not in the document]
cryptoAtwill: added back

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -104,52 +104,6 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         IGateway(s.ipcGatewayAddr).kill();
     }
 
-    /// @notice methods that allows a validator to submit a checkpoint (batch of messages) and vote for it with it's own voting power.
-    /// @param checkpoint - the batch messages data
-    function submitCheckpoint(BottomUpCheckpoint calldata checkpoint) external {
```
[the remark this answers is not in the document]
cryptoAtwill: added back

src/gateway/GatewayRouterFacet.sol
cryptoAtwill: I added this into the interface already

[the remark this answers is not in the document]
cryptoAtwill: Right, I think this is the mechanism of the top down checkpointing in `fendermint`, not sure if it suitable to place the docs here.

But long story short, membership information is pulled from the parent `ipc-agent` or `node` in `fendermint`. Then `fendermint` call this function implicitly.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -15,6 +15,10 @@ struct GatewayActorStorage {
     /// @notice a mapping of block number to cross messages
     /// SubnetID => blockNumber => messages
     mapping(bytes32 => mapping(uint256 => CrossMsg[])) topDownMsgs;
+    /// @notice The parent finalities. Key is the block number, value is the finality struct.
+    mapping(uint256 => ParentFinality) finalitiesMap;
```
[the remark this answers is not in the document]
cryptoAtwill: Not really. We just need the block height to parent finality.

src/gateway/GatewayRouterFacet.sol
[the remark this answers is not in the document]
cryptoAtwill: For top down parent finality, we dont really need mutisig.

adlrocha (approved):
Thank you for the changes, :ship:

PR #195, opened 2023-08-30
Add install eth-abi to make test

cryptoAtwill:
Some unit tests require `eth-abi` python module to work. Previously `make test` assumes `eth-abi` has already been installed, now we add install `eth-abi` into `make install-dev`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ build:
 	forge build
 
 test:
-    forge test -vvv --ffi
+	forge test -vvv --ffi
 
 install-dev: install-npm-package install-eth-abi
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -33,8 +33,13 @@ format:
...
 
-test:
-	forge test -vvv --ffi
+test: install-eth-abi forge-test
+
+forge-test:
+    forge test -vvv --ffi
+
+install-eth-abi:
+    pip install eth_abi
```
dnkolegov, line 42: It works only if `pip` has been already installed. So the suggestion above should install pip, eth_abi and all other dependencies for development.

dnkolegov (approved):
Thank you! Look good!

adlrocha (approved):

PR #196, opened 2023-08-30
Fix sendCrossMessage and execute methods

dnkolegov:
This PR suggests important fixes and improvements for crossnet messaging.

- I believe there were several critical bugs in `execute` call. The tests and the contracts worked because nobody used crossnet messaging with wrapped messages and crossnet calls.
- I derived logic for `wrapped` messages from the code, which may be incorrect.
- Changed protocol for `fee` and `value` in `sendCrossMessage`.
- It fixes several critical name collisions in the tests that might lead to unexpected behavior in tests in the future. Again, many tests before were incorrect
- Added two tests that use crossnet calls (simulation for ERC20) and crossnet-messaging with value.
- Added an example contract for how to use `sendCrossMessage`


src/gateway/GatewayMessengerFacet.sol
```diff
@@ -17,15 +17,10 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
...
+    /// @notice sends an arbitrary cross message from the local subnet to the destination subnet
     /// @param crossMsg - message to send
     function sendCrossMessage(CrossMsg calldata crossMsg) external payable hasFee {
-        // There can be many semantics of the (rawAddress, msg.sender) pairs.
-        // It depends on who is allowed to call sendCrossMessage method and what we want to get as a result.
-        // They can be equal, we can propagate the real sender address only or both.
-        // We are going to use the simplest implementation for now and define the appropriate interpretation later
-        // based on the business requirements.
-        if (crossMsg.message.value != msg.value) {
+        if (crossMsg.message.value != msg.value - s.crossMsgFee) {
```
adlrocha, line 23: We should document that the `msg.value` expects whatever wants to be sent in the `crossMsg.value` plus the fee so it is clear for users.

src/lib/CrossMsgHelper.sol
```diff
@@ -87,7 +87,7 @@ library CrossMsgHelper {
             params = abi.encode(crossMsg);
         }
 
-        bytes memory data = abi.encodeWithSelector(crossMsg.message.method, params);
+        bytes memory data = bytes.concat(crossMsg.message.method, params);
```
adlrocha, line 90: I know you explained it to me, but just to be clear, Ethereum transactions concatenate the method and its arguments as bytes in the `data` when a transaction call is done between contracts?

test/GatewayDiamond.t.sol
```diff
@@ -1768,6 +1774,121 @@ contract GatewayDiamondDeploymentTest is StdInvariant, Test {
...
+
+        SubnetID memory destinationSubnet = gwGetter.getNetworkName().createSubnetId(receiver);
+
+        GatewayDiamond destinationGWDiamond = createGWForSubnet(destinationSubnet);
+
+        vm.startPrank(dstTokenOwner);
+        dstToken.transfer(address(destinationGWDiamond), targetTokenValue);
+        vm.stopPrank();
+
+        CrossMsg memory crossMsg = CrossMsg({
```
adlrocha, line 1857: Really nice test, thank you for this! We were really missing it.

adlrocha (approved):
Amazing job, @dnkolegov, I think all your changes make sense, and they were definitely critical issues. :raised_hands:

src/gateway/GatewayMessengerFacet.sol
dnkolegov: Agree. What is the best place to document it?

src/lib/CrossMsgHelper.sol
dnkolegov: `abi.encodeWithSelector` specification suggests calling it with real types. 

For example,
`abi.encodeWithSelector(bytes4(keccak256("transfer(address,uint256)")), addr, value)`. But it is incorrect to call it with `params = abi.encode(user, targetTokenValue)`, because it suggests that the `transfer` function signature is `transfer(bytes)`.

Under the hood, `abi.encodeWithSelector` does the same. It concatenates the selector with encoded params, but params must have the correct signature. 

Now, we do exactly the same.

src/gateway/GatewayMessengerFacet.sol
dnkolegov: Fixed

PR #199, opened 2023-09-05
IPC-198: Worker address as a string and remove genesis

### LINKED ISSUE #198: Remove `genesis` and make `workerAddr` a `string`
Related: https://github.com/consensus-shipyard/fendermint/issues/152

For the integration of reconfiguration and genesis generation for subnets in Fendermint, we need some changes in the solidity contracts. Mainly, 
- Make the worker address a `string` so we can provide a public key or any other type of data required to identify the validator. 
- Remove the genesis block. Subnet genesis generation will be performed deterministically from the subnet actor info, so there's no need to store it on-chain.

adlrocha:
Closes #198

### DIFF
--- a/src/lib/LibSubnetActorStorage.sol
+++ b/src/lib/LibSubnetActorStorage.sol
@@ -22,11 +22,9 @@ struct SubnetActorStorage {
     /// @notice validator address to validator net address
     mapping(address => string) validatorNetAddresses;
     /// @notice validator address to validator worker address
-    mapping(address => FvmAddress) validatorWorkerAddresses;
+    mapping(address => string) validatorWorkerKeys;
     /// @notice contains all committed bottom-up checkpoint at specific epoch
     mapping(uint64 => BottomUpCheckpoint) committedCheckpoints;
-    /// @notice genesis block
-    bytes genesis;
     /// @notice Total collateral currently deposited in the GW from the subnet
     uint256 totalStake;
     /// @notice Minimal activation collateral
--- a/src/structs/Validator.sol
+++ b/src/structs/Validator.sol
@@ -6,7 +6,7 @@ import {FvmAddress} from "../structs/FvmAddress.sol";
 struct ValidatorInfo {
     address addr;
     uint256 weight;
-    FvmAddress workerAddr;
+    string workerKey;
     string netAddresses;
 }
 
--- a/src/subnet/SubnetActorGetterFacet.sol
+++ b/src/subnet/SubnetActorGetterFacet.sol
@@ -79,10 +79,6 @@ contract SubnetActorGetterFacet {
         return s.bottomUpCheckPeriod;
     }
 
-    function genesis() external view returns (bytes memory) {
-        return s.genesis;
-    }
-
     function majorityPercentage() external view returns (uint64) {
         return LibVoting.majorityPercentage();
     }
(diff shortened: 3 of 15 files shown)


src/lib/LibSubnetActorStorage.sol
```diff
@@ -22,11 +22,9 @@ struct SubnetActorStorage {
     /// @notice validator address to validator net address
     mapping(address => string) validatorNetAddresses;
     /// @notice validator address to validator worker address
-    mapping(address => FvmAddress) validatorWorkerAddresses;
+    mapping(address => string) validatorWorkerKeys;
```
dnkolegov, line 25: Why Keys, not Addresses?

src/structs/Validator.sol
```diff
@@ -6,7 +6,7 @@ import {FvmAddress} from "../structs/FvmAddress.sol";
 struct ValidatorInfo {
     address addr;
     uint256 weight;
-    FvmAddress workerAddr;
+    string workerKey;
```
dnkolegov, line 9: If it is `workerKey` and not `workerAddr`, then it would be good to have a comment explaining that

src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -118,8 +114,8 @@ contract SubnetActorGetterFacet {
 
     /// @notice get validator worker address
```
dnkolegov, line 115: The comment says `worker address` not `worker key`

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -36,7 +36,7 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
 
     /// @notice method that allows a validator to join the subnet
     /// @param netAddr - the network address of the validator
-    function join(string calldata netAddr, FvmAddress calldata workerAddr) external payable notKilled {
+    function join(string calldata netAddr, string calldata workerKey) external payable notKilled {
```
dnkolegov, line 39: `workerKey` is missing in the description of the func

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -217,12 +217,12 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
     }
 
-    function setValidatorWorkerAddr(FvmAddress calldata newWorkerAddr) external {
+    function setValidatorWorkerKey(string calldata newWorkerKey) external {
         address validator = msg.sender;
         if (!s.validators.contains(validator)) {
             revert NotValidator();
         }
-        s.validatorWorkerAddresses[validator] = newWorkerAddr;
+        s.validatorWorkerKeys[validator] = newWorkerKey;
```
dnkolegov, line 225: Do we need to add validation here before writing to the map?
An empty string is allowed?

dnkolegov (commented):
LGTM!
But I did not understand why `key` is used instead of `addr`

adlrocha:
I think this PR has been superseded by the new `data` field included in the validator set in this PR: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/212#discussion_r1342535900

@dnkolegov, @cryptoAtwill, feel free to re-open if this is not the case.

PR #200, opened 2023-09-12
Membership

dnkolegov:
This PR addresses https://github.com/consensus-shipyard/fendermint/issues/245 and the membership mechanism.

There was an initial implementation of `ValidatorSet` type that allows us to compute and set weights for a validator set. But it was impossible to return the current membership using that type as https://github.com/consensus-shipyard/fendermint/issues/245 requires.

I combined both implementations because, at present, it is not clear what computations and interfaces we need for membership.

I applied the terms and implementation ideas we used to implement in Eudico.

@cryptoAtwill, please check that it is aligned with your previous ideas for `ValidatorSet`.
@aakoshh, please check that the introduced interfaces can be used from Fendermint.


src/gateway/GatewayManagerFacet.sol
```diff
@@ -161,4 +162,9 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
 
         LibGateway.commitBottomUpMsg(crossMsg);
     }
+
+    // /// @notice set a new membership directly
```
aakoshh, line 166:
```suggestion
    /// @notice set a new membership directly
```

dnkolegov: Thank you!

src/lib/LibGateway.sol
```diff
@@ -73,33 +74,39 @@ library LibGateway {
 
         GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
 
-        // invalidate the previous validator set
-        ++s.validatorNonce;
+        // invalidate the previous configuration number for membership
+        ++s.configurationNumber;
```
aakoshh, line 78: This is not going to work IMO, the `configurationNumber` has to be passed in explicitly from the outside. 

This is in the child subnet, right? The way I think this works is as follows:

1. Say the current configuration number is 1, something the parent and the child both agree on. 
2. Alice wants to become a validator in block A; this change creates configuration 2 in the parent. 
3. Bob wants to become a validator in block B; this change creates configuration 3 in the parent. 
4. Say in the child subnet, because the finality provider does polling, when we look at the parent chain, we already see block B as final (and thus A as well), so we propose that we finalize B, skipping A. 
5. @cryptoAtwill 's mechanism gets the validator set a B, and all cross messages since the last finalized height, but it does not get in-between states, so skips A. 
6. We call `setMembership(validators at B)` to store the validators. This is configuration 3, however the contract assigns configuration number 2 instead. 
7. Later when we create a checkpoint, I call `getConfigurationNumber` and see 2, so I set it as the next adopted validator set and send it to the parent. 
8. The parent remembers that the next checkpoint will be signed by validators in configuration 2. 
9. Later on, the parent will reject Bob's signature, who is part of configuration 3, not 2.

src/lib/LibGateway.sol
```diff
@@ -213,4 +220,11 @@ library LibGateway {
         subnet = s.subnets[subnetId.toHash()];
         found = !subnet.id.isEmpty();
     }
+
+    /// @notice returns the membership corresponding to the given epoch
+    /// @param epoch the epoch number, the membership of which we want to get
+    function getMembership(uint64 epoch) internal view returns (Membership storage) {
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+        return s.membership[epoch];
```
aakoshh, line 228: I don't understand, is this a different array than [here](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/200/files#diff-c33605fbd68a5aeb1782c3068dcb37d917cbc29d47cc2b38c3a06fd0543bb25eR169)?

src/lib/LibGateway.sol
```diff
@@ -73,33 +74,39 @@ library LibGateway {
...
             }
 
             unchecked {
-                ++validatorIndex;
+                ++i;
             }
         }
 
-        s.totalWeight = totalValidatorsWeight;
+        s.currentTotalWeight = totalValidatorsWeight;
```
aakoshh, line 108: When is `setMembership` called? I see it's `internal`, but just want to make sure that this does not get called when we finalize a new block height, because it seems to adopt them immediately, which we do not want. We want to store the new membership, but not adopt them until the end of the checkpointing period, so the `s.currentTotalWeight` and the `s.configurationNumber` should not change.

src/structs/Validator.sol
```diff
@@ -14,3 +14,14 @@ struct ValidatorSet {
...
+
+struct Validator {
+    uint256 weight;
+    FvmAddress addr;
+}
+
+struct Membership {
+    Validator[] validators;
+    uint64 configurationNumber;
+    uint256 totalWeight;
```
aakoshh, line 26: If `totalWeight` is here, why do you store `currentTotalWeight` separately?

src/lib/LibGateway.sol
aakoshh: Why is it indexed by block height here and not configuration number?

dnkolegov: The same. I updated the getter.

src/structs/Validator.sol
dnkolegov: `totalWeight ` was stored before. We will remove it if all agree with the current approach.
From what I have read in the issues, it seems that we should store `totalWeight` for each membership.

src/lib/LibGateway.sol
dnkolegov: `setMembership` is also external, accessible for SYSTEM ACTOR only.

See also this. 
https://github.com/consensus-shipyard/ipc-solidity-actors/blob/34f3bf69e44c8b9da167dd213a512331d11a405b/src/gateway/GatewayRouterFacet.sol#L41

aakoshh (commented):
Thanks! 

As I commented it's not clear to me when to call this method (even if the host _should_ call it at all), because a new membership seems to take immediate effect. There also seems to be some ambiguity whether membership is tracked by height or configuration number. 

Ideally the configuration number would be a parameter, not a computed value, and there would be two slots:
1. The current `Membership`, which does not change until the end of the checkpoint period
2. The next `Membership`, which can change multiple times during the checkpoint period but does not take effect until we call the method in https://github.com/consensus-shipyard/fendermint/issues/246

With those, the checkpoint period goes like this:
1. We receive any number of new proposed `Membership` from the parent in top-down messages. _Some_ of these we finalize and pass to the Gateway, which should _overwrite the next_ membership, but _leave current_ alone. 
2. At the end of the epoch we call the transition method which _overwrites current with next_ (unless next is empty) and returns the new current one.

Besides this it is useful to remember the configuration history. Probably the one which will ask the Gateway to remember the list of unsigned/unrelayed checkpoints, each showing both the current and the next configuration number, so that validators who restart their nodes can tell whether they should have signed it or not, to re-broadcast signatures or not.

src/structs/Validator.sol
aakoshh: Yes, I think `totalWeight` should stay and `currentTotalWeight` is redundant (may be an optimisation, but I'd expect you need to access the Membership to deal with signatures anyway).

aakoshh: `totalWeight` is useful when you have a multiple checkpoints in a list and you are collecting signatures: you need to know when you reached quorum, and they can be based on an earlier validator set.

src/lib/LibGateway.sol
adlrocha: ^+1 to this comment. `set_membership` is called by the child subnet when Fendermint adopts a new configuration. So it should pass it all including the configuration number assigned in the parent. This is just the on-chain view of the current configuration in the network. That way contracts can leverage it for their operation.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -28,7 +29,13 @@ struct GatewayActorStorage {
     mapping(uint64 => BottomUpCheckpoint) bottomUpCheckpoints;
     /// @notice List of validators and how many votes of the total each validator has for top-down messages
     // validatorNonce => validator fvm address => weight
-    mapping(uint256 => mapping(bytes32 => uint256)) validatorSet;
+    mapping(uint256 => mapping(bytes32 => uint256)) validatorSetWeights;
+    /// @notice The current configuration number.
+    uint64 configurationNumber;
+    /// @notice List of validators and their weights for each configuration number
+    mapping(uint64 => Membership) membership;
```
adlrocha, line 36: I really don't think we should keep the history of all the configuration in the gateway. This information is only used in child subnets to determine the current membership being used by Fendermint.

If one wants to check what was the configuration at a certain epoch, it just needs to go to the specific epoch and inspect the state then (contracts won't need this historical information, and if they do, they should index it themselves, it is not needed for the core operation of our protocol)

src/lib/LibGatewayActorStorage.sol
```diff
@@ -28,7 +29,13 @@ struct GatewayActorStorage {
...
+    /// @notice total votes of all validators in the current membership
+    uint256 currentTotalWeight;
```
adlrocha, line 38: This field seems redundant. It should be included as part of the membership info.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -28,7 +29,13 @@ struct GatewayActorStorage {
     mapping(uint64 => BottomUpCheckpoint) bottomUpCheckpoints;
     /// @notice List of validators and how many votes of the total each validator has for top-down messages
     // validatorNonce => validator fvm address => weight
-    mapping(uint256 => mapping(bytes32 => uint256)) validatorSet;
+    mapping(uint256 => mapping(bytes32 => uint256)) validatorSetWeights;
+    /// @notice The current configuration number.
+    uint64 configurationNumber;
```
adlrocha, line 34: We don't need this field, it should be included as part of the current membership.

adlrocha (commented):
I agree with @aakoshh comments.
- The subnet actor keeps track of all of the configurations in the parent, and is the one responsible for increasing the config number when there's a change in membership.
- The `set_membership` in the gateway is used in child subnets to reflect the current configuration adopted (no need to store the historical info, it is implicitly stored as part of the history of the chain). 

(And maybe this is part of some other issue), but the subnet actor may need to also reflect in the parent the current configuration adopted in the child for the checkpoint multisig verification. It will be notified through a bottom-up checkpoint of what is the current configuration number adopted in the child.

src/lib/LibGateway.sol
cryptoAtwill: it is because index by height should have served the requirements for top down, might need some changes for bottom up.

dnkolegov:
@adlrocha 

> The subnet actor keeps track of all of the configurations in the parent, and is the one responsible for increasing the config number when there's a change in membership.

At present, membership in a SubnetActor is static (it is equal to all validators in the subnet) and we do not have the means to change membership there. Correct? How are we going to address that? Or each client should write their own function to update/rotate/support membership?

src/lib/LibGateway.sol
aakoshh: I still don't get it what you are saying: what should it be, block height, or config number? What has been updated?

[Here](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/200/files#diff-565acf6db4298fb1595e6bca86a27eccfab40e5a1bd48d4a278282c11567a85bR84) the record is inerted _by configuration number_ and you are saying it's the same array, so how can it be indexed by epoch, which evolves independently?

aakoshh (approved):
LGTM!

src/gateway/GatewayRouterFacet.sol
```diff
@@ -33,12 +33,12 @@ contract GatewayRouterFacet is GatewayActorModifiers {
     /// @notice commit the ipc parent finality into storage
     function commitParentFinality(
         ParentFinality calldata finality,
+        uint64 n,
```
cryptoAtwill, line 36: what does `n` mean? It's the configuration number? Maybe we can use configuration number instead or add some comment to this?

src/lib/LibGateway.sol
```diff
@@ -63,43 +64,58 @@ library LibGateway {
...
+    function newMembership(uint64 n, FvmAddress[] memory validators, uint256[] memory weights) internal {
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+
         if (validators.length != weights.length) {
             revert ValidatorsAndWeightsLengthMismatch();
         }
 
-        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+        if (n <= s.lastMembership.configurationNumber) {
+            return;
```
cryptoAtwill, line 80: I'm not sure if we should just internalise this or we should have reverted the error. This is called implicitly, if this happens, could mean something is wrong. If we don't revert or emit event or return some status, when debugging, there is really no way to know what went wrong. Maybe a revert with error can help surface out the error?

src/lib/LibGateway.sol
```diff
@@ -63,43 +64,58 @@ library LibGateway {
...
         }
+        s.lastMembership.totalWeight = totalValidatorsWeight;
+    }
 
-        s.totalWeight = totalValidatorsWeight;
+    /// @notice update the membership of the child subnet and returns it
+    function updateMembership() internal returns (Membership memory) {
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+        s.currentMembership = s.lastMembership;
+        return s.currentMembership;
```
cryptoAtwill, line 118: Just curious, will the compiler optimize this or which might be more gas efficient compared to:
```
Membership memory m = s.latestMemberhsip;
s.currentMembership = m;
return m;
```
Will `return s.currentMembership;` trigger another load or compiler will just read in memory.

src/gateway/GatewayRouterFacet.sol
dnkolegov: n is a configuration number

src/lib/LibGateway.sol
dnkolegov: I did not check. If you think this one it more efficient we can use it

src/lib/LibGateway.sol
dnkolegov: If we revert here then `commitParentFinality` will be also reverted. But you need to call it with the same n.

cc @aakoshh

src/lib/LibGatewayActorStorage.sol
```diff
@@ -28,7 +29,13 @@ struct GatewayActorStorage {
     mapping(uint64 => BottomUpCheckpoint) bottomUpCheckpoints;
     /// @notice List of validators and how many votes of the total each validator has for top-down messages
     // validatorNonce => validator fvm address => weight
```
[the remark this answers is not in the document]
dnkolegov: It better describes membership from my perspective

src/lib/LibGateway.sol
cryptoAtwill: no strong preference here, just curious.

src/lib/LibGateway.sol
cryptoAtwill: Purely from the debugging point of view, if we return here, we cannot distinguish between if the configuration number is committed or just returned. It wouldn't be very useful for debugging if sth goes wrong. We could add an event when new configuration number is committed to distinguish.

If we just return here, this means the new parent finality will be committed, but the configuration number is not the one associated with the parent finality. At this point, I'm not sure how big an impact it would be or how would we end up in this situation. But gut feeling says the two updates should be atomic, I could be wrong.

dnkolegov: In this case there should not be an update for membership. Nothing to update. The membership is the same

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/lib/LibGatewayActorStorage.sol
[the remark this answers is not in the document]
dnkolegov: Fixed

src/lib/LibGateway.sol
dnkolegov: That one does not work.

```
Unimplemented feature (/solidity/libsolidity/codegen/ArrayUtils.cpp:228):Copying of type struct Validator memory[] memory to storage not yet supported.
```

src/lib/LibGateway.sol
dnkolegov: @cryptoAtwill and @aakoshh please check https://github.com/consensus-shipyard/ipc-solidity-actors/pull/200/commits/a6e1ba12bdd8d8784edb1a8d3f004b73ef64c9d8

I hope I have added the suggestions of you both.

src/gateway/GatewayGetterFacet.sol
```diff
@@ -156,4 +153,34 @@ contract GatewayGetterFacet {
...
+        return s.lastMembership;
+    }
+
+    /// @notice get the last configuration number received from the parent
+    function getLastConfigurationNumber() external view returns (uint64) {
+        return s.lastMembership.configurationNumber;
+    }
+
+    /// @notice get the last total weight for the membership received from the parent
+    function lastTotalWeight() public view returns (uint256) {
```
aakoshh, line 168: Minor but I wonder why the other to are `getFooBar` while this is just `fooBar`.

src/gateway/GatewayManagerFacet.sol
```diff
@@ -161,4 +162,18 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
...
+    /// @notice set a new membership according to the configuration received from the parent
+    function newMembership(
+        uint64 n,
+        FvmAddress[] memory validators,
+        uint256[] memory weights
+    ) external systemActorOnly {
+        LibGateway.newMembership({n: n, validators: validators, weights: weights});
+    }
+
+    /// @notice updates the membership of the child subnet and returns it
```
aakoshh, line 175:
```suggestion
    /// @notice updates the current membership of the child subnet to the last received one and returns the new value
```

src/lib/LibGateway.sol
```diff
@@ -23,6 +24,9 @@ library LibGateway {
     using CrossMsgHelper for CrossMsg;
     using CheckpointHelper for BottomUpCheckpoint;
 
+    event NewMembership(uint64 n, FvmAddress[] validators, uint256[] weights);
+    event MembershipUpdated(uint64 n, Validator[] validators, uint256 totalWeight);
```
aakoshh, line 28: Is there any reason why these to events aren't isomorhpic?

src/lib/LibGateway.sol
```diff
@@ -63,43 +67,70 @@ library LibGateway {
...
+        for (uint256 i = 0; i < len; ) {
+            if (!FvmAddressHelper.equal(validators[i], zero)) {
+                uint256 validatorWeight = weights[i];
 
                 if (validatorWeight == 0) {
                     revert ValidatorWeightIsZero();
                 }
 
-                s.validatorSet[s.validatorNonce][validators[validatorIndex].toHash()] = validatorWeight;
+                s.validatorSetWeights[n][validators[i].toHash()] = validatorWeight;
```
aakoshh, line 113: I thought we agreed not to store history for now?

src/lib/LibGateway.sol
```diff
@@ -63,43 +67,70 @@ library LibGateway {
...
 
-    /// @notice set up the top-down validators and their voting power
+    /// @notice set the next membership
+    /// @dev It is called by the child subnet when a new configuration is adopted.
+    /// @param n - configuration number
     /// @param validators - list of validator addresses
     /// @param weights - list of validators voting powers
-    function setMembership(FvmAddress[] memory validators, uint256[] memory weights) internal {
+    function newMembership(uint64 n, FvmAddress[] memory validators, uint256[] memory weights) internal {
+        emit NewMembership({n: n, validators: validators, weights: weights});
```
aakoshh, line 76: I reckon this should only be emitted if it's actually new. I'd do it at the end of the method, and then the format can be the same as the one in `updateMembership` too.

aakoshh (commented):
Thanks, almost there! 

My understanding from @adlrocha comments were that he also didn't want to keep the validator history in the gateway.

My other remark is to only emit `NewMemberhip` events after we verified that it's indeed new.

src/lib/LibGateway.sol
dnkolegov: You and I did, but not @cryptoAtwill.

@cryptoAtwill could you add here your concerns? Thank you.

src/lib/LibGateway.sol
dnkolegov: The point of this event is debugging in case we ignore or reject something. Maybe I should name it like `ReceivedMembership`? But the goal of this event is to show that we call newMembership method.

src/lib/LibGateway.sol
dnkolegov: Just to save more gas and not create a new data structure. Because in the `updateMembership` function we do not have access to the `weights` array.

src/lib/LibGateway.sol
dnkolegov: renamed to `MembershipReceived`

src/lib/LibGateway.sol
```diff
@@ -63,43 +67,70 @@ library LibGateway {
...
+        s.lastMembership.totalWeight = totalValidatorsWeight;
+    }
 
-        s.totalWeight = totalValidatorsWeight;
+    /// @notice update the membership of the child subnet and returns it
+    function updateMembership() internal returns (Membership memory) {
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+        s.currentMembership = s.lastMembership;
+        Membership memory mb = s.currentMembership;
+        emit MembershipUpdated({n: mb.configurationNumber, validators: mb.validators, totalWeight: mb.totalWeight});
```
aakoshh, line 132: This should ideally only be emitted if there was a change IMO.

dnkolegov: Correct. `MembershipUpdated` is emitted only if the membership has been changed.
`MembershipReceived` is emitted each time `newMembership` is called.

Or do you mean that even if `updateMembership` is called, that does not mean the membership will be changed?

src/gateway/GatewayGetterFacet.sol
dnkolegov: Because we do not have a code style.
Fixed

src/lib/LibGateway.sol
dnkolegov: Fixed. Good catch.

PR #201, opened 2023-09-14
Remove min params

dnkolegov:
This PR removes min values for collateral and submission period according to the https://github.com/consensus-shipyard/fendermint/pull/267#issuecomment-1718091638

### DIFF
--- a/src/GatewayDiamond.sol
+++ b/src/GatewayDiamond.sol
@@ -3,7 +3,7 @@ pragma solidity 0.8.19;
 
 import {GatewayActorStorage} from "./lib/LibGatewayActorStorage.sol";
 import {IDiamond} from "./interfaces/IDiamond.sol";
-import {InvalidCollateral} from "./errors/IPCErrors.sol";
+import {InvalidCollateral, InvalidSubmissionPeriod} from "./errors/IPCErrors.sol";
 import {LibDiamond} from "./lib/LibDiamond.sol";
 import {LibVoting} from "./lib/LibVoting.sol";
 import {SubnetID} from "./structs/Subnet.sol";
@@ -27,19 +27,23 @@ contract GatewayDiamond {
     }
 
     constructor(IDiamond.FacetCut[] memory _diamondCut, ConstructorParams memory params) {
-        LibDiamond.setContractOwner(msg.sender);
-        LibDiamond.diamondCut({_diamondCut: _diamondCut, _init: address(0), _calldata: new bytes(0)});
-
         if (params.minCollateral == 0) {
             revert InvalidCollateral();
         }
+        // topDownCheckPeriod can be equal 0, since validators can propose anything they want.
+        // The bottomUpCheckPeriod should be non-zero for now.
+        if (params.bottomUpCheckPeriod == 0) {
+            revert InvalidSubmissionPeriod();
+        }
+
+        LibDiamond.setContractOwner(msg.sender);
+        LibDiamond.diamondCut({_diamondCut: _diamondCut, _init: address(0), _calldata: new bytes(0)});
 
         s.networkName = params.networkName;
         s.minStake = params.minCollateral;
         s.bottomUpCheckPeriod = params.bottomUpCheckPeriod;
         s.topDownCheckPeriod = params.topDownCheckPeriod;
         s.crossMsgFee = params.msgFee;
-        s.minStake = params.minCollateral;
 
         // the root doesn't need to be explicitly initialized
         if (s.networkName.isRoot()) {
--- a/src/GatewayDiamond.sol
+++ b/src/GatewayDiamond.sol
@@ -16,11 +16,7 @@ contract GatewayDiamond {
 
     using SubnetIDHelper for SubnetID;
 
-    // @notice uint8 constant MIN_CHECKPOINT_PERIOD = 10;
     uint256 public constant MIN_COLLATERAL_AMOUNT = 1 ether;
-
-    /// @notice minimum checkpoint period. Values get clamped to this
-    uint8 public constant MIN_CHECKPOINT_PERIOD = 10;
 
     struct ConstructorParams {
         SubnetID networkName;
@@ -36,12 +32,8 @@ contract GatewayDiamond {
 
         s.networkName = params.networkName;
         s.minStake = MIN_COLLATERAL_AMOUNT;
-        s.bottomUpCheckPeriod = params.bottomUpCheckPeriod < MIN_CHECKPOINT_PERIOD
-            ? MIN_CHECKPOINT_PERIOD
-            : params.bottomUpCheckPeriod;
-        s.topDownCheckPeriod = params.topDownCheckPeriod < MIN_CHECKPOINT_PERIOD
-            ? MIN_CHECKPOINT_PERIOD
-            : params.topDownCheckPeriod;
+        s.bottomUpCheckPeriod = params.bottomUpCheckPeriod;
+        s.topDownCheckPeriod = params.topDownCheckPeriod;
         s.crossMsgFee = params.msgFee;
 
         // the root doesn't need to be explicitly initialized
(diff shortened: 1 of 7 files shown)


src/SubnetActorDiamond.sol
```diff
@@ -46,15 +40,17 @@ contract SubnetActorDiamond {
         if (params.ipcGatewayAddr == address(0)) {
             revert GatewayCannotBeZero();
         }
+        if (params.topDownCheckPeriod == 0 || params.bottomUpCheckPeriod == 0) {
```
aakoshh, line 43: I wouldn't necessarily consider `topDownCheckPeriod = 0` an invalid value. Currently in @cryptoAtwill implementation there is no fixed period, validators can propose anything they want (limits might be put on the volume of messages in the range since the last finalized height). If anything I would reserve 0 for "don't care".

The `bottomUpCheckPeriod` should be non-zero for now :+1:

src/GatewayDiamond.sol
```diff
@@ -34,15 +30,16 @@ contract GatewayDiamond {
         LibDiamond.setContractOwner(msg.sender);
         LibDiamond.diamondCut({_diamondCut: _diamondCut, _init: address(0), _calldata: new bytes(0)});
 
+        if (params.minCollateral == 0) {
+            revert InvalidCollateral();
+        }
```
aakoshh, line 35: Why does `SubnetActorDiamond.sol` check more conditions than `GatewayDiamond.sol`? By the looks of it the same fields are present in both.

src/GatewayDiamond.sol
```diff
@@ -16,16 +17,11 @@ contract GatewayDiamond {
...
-    uint256 public constant MIN_COLLATERAL_AMOUNT = 1 ether;
-
-    /// @notice minimum checkpoint period. Values get clamped to this
-    uint8 public constant MIN_CHECKPOINT_PERIOD = 10;
-
     struct ConstructorParams {
         SubnetID networkName;
         uint64 bottomUpCheckPeriod;
         uint64 topDownCheckPeriod;
+        uint256 minCollateral;
```
aakoshh, line 24: Why did you add another constructor argument _now_? 

So this is the gateway; what does `minCollateral` apply to? 
* Is it for the validators of _this_ subnet? In that case, why isn't it enforced in the parent `SubnetActor`?
* Is it for the validators of the child subnets? In that case, how can we come up with a value that applies to everything? Why aren't they defined on individual `SubnetActor` instances when they are created?

src/GatewayDiamond.sol
```diff
@@ -34,15 +30,16 @@ contract GatewayDiamond {
...
-            ? MIN_CHECKPOINT_PERIOD
-            : params.bottomUpCheckPeriod;
-        s.topDownCheckPeriod = params.topDownCheckPeriod < MIN_CHECKPOINT_PERIOD
-            ? MIN_CHECKPOINT_PERIOD
-            : params.topDownCheckPeriod;
+        s.minStake = params.minCollateral;
+        s.bottomUpCheckPeriod = params.bottomUpCheckPeriod;
+        s.topDownCheckPeriod = params.topDownCheckPeriod;
         s.crossMsgFee = params.msgFee;
+        s.minStake = params.minCollateral;
```
aakoshh, line 42: Already assigned on line 38

src/SubnetActorDiamond.sol
dnkolegov: Fixed

aakoshh (commented):
Thank you for these changes! 

I'm not clear on why the constructor arguments have been extended with a new field all of a sudden, and I think we can potentially relax the checks even more and accept some zeroes.

src/SubnetActorDiamond.sol
```diff
@@ -46,15 +40,17 @@ contract SubnetActorDiamond {
...
+            revert InvalidSubmissionPeriod();
+        }
+        if (params.minActivationCollateral == 0) {
+            revert InvalidCollateral();
+        }
```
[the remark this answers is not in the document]
dnkolegov: There is a logic around `minCollateral` - https://github.com/consensus-shipyard/ipc-solidity-actors/blob/34f3bf69e44c8b9da167dd213a512331d11a405b/src/subnet/SubnetActorManagerFacet.sol#L47-L68

From this logic, it must be greater than zero.

src/GatewayDiamond.sol
dnkolegov: That was a bug. Now fixed

src/GatewayDiamond.sol
dnkolegov: I added it because before we had this https://github.com/consensus-shipyard/ipc-solidity-actors/blob/main/src/GatewayDiamond.sol#L38

`minStake` is used in the GatewayManagerFacet.
- https://github.com/consensus-shipyard/ipc-solidity-actors/blob/34f3bf69e44c8b9da167dd213a512331d11a405b/src/gateway/GatewayManagerFacet.sol#L36
- https://github.com/consensus-shipyard/ipc-solidity-actors/blob/34f3bf69e44c8b9da167dd213a512331d11a405b/src/gateway/GatewayManagerFacet.sol#L73
- https://github.com/consensus-shipyard/ipc-solidity-actors/blob/34f3bf69e44c8b9da167dd213a512331d11a405b/src/gateway/GatewayManagerFacet.sol#L96

src/GatewayDiamond.sol
dnkolegov: Fixed

src/GatewayDiamond.sol
aakoshh: Although I can see the value in a subnet-creation-policy that can put some restrictions on the number and quality of subnets created.

aakoshh: Okay thanks, created https://github.com/consensus-shipyard/fendermint/issues/269 to add the config counterpart.

src/SubnetActorDiamond.sol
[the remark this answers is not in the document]
aakoshh: Okay, I can imagine a non-zero total stake avoids some division-by-zero somewhere.

Why is an individual validator above compared to the subnet activation collateral?

```
if (s.stake[validator] >= s.minActivationCollateral) { 
```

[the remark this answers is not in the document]
dnkolegov: That's the logic we have had from the beginning. So this is the default policy for our subnets.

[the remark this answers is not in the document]
aakoshh: It doesn't sound right, because as far as I know the subnet collateral can fall behind the minimum and cause the subnet to be deactivated. However if every single validator has to put up at least as much as the miniumum required collateral, then falling behind the threshold means all validators have abandoned the subnet.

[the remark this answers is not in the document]
dnkolegov: @aakoshh We were trying to fix it https://github.com/consensus-shipyard/ipc-solidity-actors/issues/71 but decided to postpone it. The reasons are in the last message.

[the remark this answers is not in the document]
dnkolegov: I reopened that issue

aakoshh (approved):
With the outstanding question of activation collateral, I think this change is ready :+1:

src/SubnetActorDiamond.sol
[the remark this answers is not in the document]
aakoshh: Thanks, I'll send my comments there.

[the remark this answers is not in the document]
adlrocha: So IIUC, the question is if `minActivationCollateral` can be zero. This is the reference implementation for the subnet actor, and I think it should be that way. The gateway has a minimum collateral already to register a subnet into the IPC hierarchy _that is independent from the one set by the subnet operators_. If a subnet decides that there is no minimum collateral required to join the subnet as a validator, as long as it collects enough to register their subnets to the gateway that should be fine. 

That being said, in Fendermint and Mir/Trantor, allowing `0` may not be the best of the ideas, as it would set the weight for the validator to `0`, which is like not having a validator.

My take is that in the current implementation, we should allow a subnet to be able to set arbitrarily low  collateral requirements for validators in the subnet but ideally not equal to zero (as it may break a few things under our PoS subnet assumptions).

PR #202, opened 2023-09-15
Fendermint [247, 248, 249]: Checkpoint signatures

dnkolegov:
This PR addresses 
* https://github.com/consensus-shipyard/fendermint/issues/247 - Checkpointing actor
* https://github.com/consensus-shipyard/fendermint/issues/248 - Checkpoint creation
* https://github.com/consensus-shipyard/fendermint/issues/249 - Checkpoint signatures


src/gateway/GatewayRouterFacet.sol
```diff
@@ -30,6 +33,10 @@ contract GatewayRouterFacet is GatewayActorModifiers {
     using FvmAddressHelper for FvmAddress;
     using StorableMsgHelper for StorableMsg;
 
+    event SignaturesUpdated(uint256 h, bytes32 pof, Membership membership, bytes signature);
+    event SignatureInvalid(uint256 h, bytes32 pof, Membership membership, bytes signature);
+    event PoFThresholdReached(uint256 h, bytes32 pof, Membership membership, uint256 threshold);
```
aakoshh, line 38: What are `h` and `pof`? If `h` is block height, shouldn't it be `uint64`? Does it cost anything to have longer field names? 

A rule of thumb was to make field names which are long lived more verbose, and only use single letters for short-lived local variables.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,52 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
     }
+
+    /// @notice checks whether the provided proof-of-finality signature for a block at height `h ` is valid and accumulates it.
+    /// @param h - height when the signature was created
+    /// @param pof - proof of finality
+    /// @param membership - the membership at height `h`
+    /// @param signature - added signature
+    function newSignature(
+        uint256 h,
+        bytes32 pof,
```
aakoshh, line 197: Is `pof` supposed to be the hash of a checkpoint? We have to define the checkpoint type, it would be very useful. 

I have to say I don't like this "proof of finality" moniker, it's not actually proving anything in most cases where it's being used.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,52 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        Membership calldata membership,
+        bytes calldata signature
+    ) external systemActorOnly {
```
aakoshh, line 200: No this is actually a normal transaction broadcasted by validators, not something we apply implicitly.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,52 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        (address signer, ECDSA.RecoverError err, ) = ECDSA.tryRecover(pof, signature);
+        if (err != ECDSA.RecoverError.NoError) {
+            revert InvalidSignature();
+        }
+
+        uint256 signersNumber = membership.validators.length;
+        uint256 signerIndex;
+        bool signerExists;
+        for (signerIndex = 0; signerIndex < signersNumber; ) {
+            if (signer == membership.validators[signerIndex].addr.extractEvmAddress().normalize()) {
```
aakoshh: Is this calculation cheaper than storing it in the first place? 

Should perhaps the signer index be an input?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,52 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+                s.bottomUpCollectedSignatures[h][pof].push(signature);
```
aakoshh, line 211: It maybe already collected.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,52 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+                break;
+            }
+
+            unchecked {
+                ++signerIndex;
+            }
+        }
+
+        if (signerExists) {
+            uint256 threshold = s.bottomUpCollectedSignaturesThreshold[h][pof];
```
aakoshh, line 223: Why the double indexing? `h` should be sufficient.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,52 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+            threshold += membership.validators[signerIndex].weight;
```
aakoshh: Does this need to be written back to the storage?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,52 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+
+            if (threshold >= membership.totalWeight) {
+                emit PoFThresholdReached({h: h, pof: pof, membership: membership, threshold: threshold});
```
aakoshh, line 227: It's enough to emit in the first time the threshold is passed, but keep adding signatures if they come in.

src/gateway/GatewayRouterFacet.sol
aakoshh: I wonder since you kept the membership history for the current one, whether its' better to copy it to the checkpoint, rather than hijack transactions in Fendermint.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -34,6 +34,12 @@ struct GatewayActorStorage {
     Membership currentMembership;
     /// @notice The last membership received from the parent and adopted
     Membership lastMembership;
+    /// @notice The signatures collected for the proof of finality at height `h`
+    mapping(uint256 => mapping(bytes32 => bytes[])) bottomUpCollectedSignatures;
+    /// @notice The signatures collected for the proof of finality at height `h`
+    mapping(uint256 => mapping(bytes32 => uint256)) bottomUpCollectedSignaturesThreshold;
+    /// @notice Bottom-up proof of finality for a block at height `h`
+    mapping(uint256 => bytes32) bottomUpProofOfFinalities;
```
aakoshh, line 42: Can we call these "bottomUpCheckpoints"? 

And let's define what these bytes are, because it's not clear to me.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -34,6 +34,12 @@ struct GatewayActorStorage {
     Membership currentMembership;
     /// @notice The last membership received from the parent and adopted
     Membership lastMembership;
+    /// @notice The signatures collected for the proof of finality at height `h`
+    mapping(uint256 => mapping(bytes32 => bytes[])) bottomUpCollectedSignatures;
+    /// @notice The signatures collected for the proof of finality at height `h`
+    mapping(uint256 => mapping(bytes32 => uint256)) bottomUpCollectedSignaturesThreshold;
```
aakoshh, line 40: As discussed there can be only one checkpoint at any height.

src/gateway/GatewayRouterFacet.sol
dnkolegov: If validators vote for `pof` that means there can be more than one `pof.  Correct?

src/gateway/GatewayRouterFacet.sol
aakoshh: I mean we can even use Merkle proofs to prove that a validator is a member of the committee, if you don't want to store the whole membership for the few outstanding checkpoints, and keep using explicit transactions from validators.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Why do we need `checkpoint` or `PoF` type exactly here instead of `bytes32`?

src/gateway/GatewayRouterFacet.sol
aakoshh: I don't think they vote for anything. They each produce a `Checkpoint` data structure independently and add it to the ledger. Subsequently they broadcast their signature over that checkpoint, and they have to check whether signatures coming in in the form of transactions are about the same checkpoint. There are no alternative checkpoints to consider and there is no voting, it's only a process to accumulate signatures (which will be the proof of finality when they reach quorum), so that they can be relayed along with the `Checkpoint` to the subnet actor in the parent.

src/gateway/GatewayRouterFacet.sol
dnkolegov: > Should perhaps the signer index be an input?

Yes, it will be more efficient! Can you provide this index in the input?

src/gateway/GatewayRouterFacet.sol
dnkolegov: I thought we had decided not to keep the membership and only the current and next (last) membership.

src/gateway/GatewayRouterFacet.sol
dnkolegov: We are in a child subnet, so, if validators A and B have different `Checkpoint`, the signatures also differ.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Then, we need to store in the state that the threshold is passed.
Or we can emit two different events - `ThresholdPassed` and `ThresholdUpdated`.

src/gateway/GatewayRouterFacet.sol
adlrocha: The current membership adopted by the child subnet should be the one signing the PoF/checkpoint (or however we are calling it) at this specific height, right? And until we adopt the proposal of just keeping a pointer to the latest membership information, the gateway should have information about the current membership in its own state, which means that we should be able, as @aakoshh suggests, to validate the signature from the existing membership information in the gateway state, right?

Even if we use the "pointer to current membership" approach, the contract should be able to verify that the one provided as an argument is the right membership and use it to make the signature verification.

TL;DR, unless we want to subsidize the gas costs for some reason (and since @aakoshh discovery of gas_price we may not even need this), this can directly be an explicit transaction, no need to protect it with systemActorOnly as no one can mess up with the system by calling the function when it shouldn't be entitle to.

src/gateway/GatewayRouterFacet.sol
aakoshh: Yeah, I think that's a good idea. What I think we want is something simple for relayers to react to: the most important event is that there is a checkpoint to relay to the parent, but in order to include as many signatures as possible, we could incentivise them to wait a little and submit when there are more. The "updated" event could be a nice way for them to know that they can rake in more, rather than polling the ledger.

src/gateway/GatewayRouterFacet.sol
aakoshh: Right, we have to think about what the SubnetActor will need in order to handle a bottom-up checkpoint, and capture that as a type. Then we can discuss what these bytes are. They can be the hash of this checkpoint, but then it's unneeded because the ledger already has this hash. 

Don't know if you _need_ a checkpoint here (I mean _here_ I think we don't need either a checkpoint or bytes), but it's the simplest way to do things.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Will add

src/lib/LibGatewayActorStorage.sol
dnkolegov: Those bytes are just the hash (sha3) of the checkpoint.

src/gateway/GatewayRouterFacet.sol
aakoshh: > The current membership adopted by the child subnet should be the one signing the PoF/checkpoint (or however we are calling it) at this specific height, right?

Not quite. We actually need the validators who prevailed _at the block height of the checkpoint_, but they may no longer be active. That's because when we created the checkpoint, we also executed the validator set transition. And there can be in theory multiple checkpoints in the pipeline awaiting their signatures. 

This is something @cryptoAtwill asked: what if these past validators don't submit their signatures? Well in that case there won't be any more checkpoints going to the parent ever, and they will never be able to get their collateral back.

src/gateway/GatewayRouterFacet.sol
aakoshh: Answered on [Slack](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1695035921212019)

src/lib/LibGatewayActorStorage.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/lib/LibGatewayActorStorage.sol
dnkolegov: Fixed. At present, this PR does not use it

src/gateway/GatewayRouterFacet.sol
cryptoAtwill: Yeah, that's the thing. In `fvm`, it's either `u64` or `i64`, but in solidity, it's `U256`. How we do this previously is just use `U256` in solidity as it's the convention. Then in rust, we convert to whatever is the preferred type. I don't feel we will run into overflow though. But feel free to suggest otherwise.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +185,59 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+            revert InvalidValidatorIndex();
+        }
+
+        address validator = membership.validators[validatorIndex].addr.extractEvmAddress().normalize();
+
+        if (validator == recoveredSignatory) {
+            s.bottomUpCollectedSignatures[height][signature] = true;
+
+            CheckpointQuorum memory quorum = s.bottomUpCheckpointQuorum[height];
+            quorum.weight += membership.validators[validatorIndex].weight;
```
cryptoAtwill, line 221: I feel passing the whole membership then just use the lookup for the `validatorIndex` is not very gas friendly. I'm thinking along the line with @aakoshh's proposal of using `MerkleProof`.

In `createCheckpoint`, it's executed implicitly, we can obtain the list of validators from `cometbft`, then we can sort the validators in a deterministic order, like by address. Then we construct the full merkle tree and get the root. In create, we just insert the root into the checkpoint. 

When perform the adding signature, we just can execute explicitly with the validator as the sender. The sender should also have the list of validators offchain and obtain its validatorIndex, i.e. position in the whole merkle tree. Then the validator can calculate the merkle proof offchain with (address, weight). Then the method signature would be something like: 
```solidity
addSignature(bytes32[] proof, u256 weight, bytes32 checkpoint, bytes signature)
```
We can validate the merkle proof using openzeppline library pretty easily to ensure this sender is part of the validator set. Then we can do the signature checks.

This way we dont really need the whole validator set, the validator index(not sure about this one) and we can execute explicitly.

src/gateway/GatewayRouterFacet.sol
aakoshh: It sounds like just a dogmatic choice that is a forever thorn in one's side of how to handle the potential conversion errors, and takes up more space, but if that's how you want it :shrug:

src/gateway/GatewayRouterFacet.sol
aakoshh: If the membership is indeed an input, then I'm sure we can ask the host to find the index. With the Merkle proofs it would definitely be required.

src/gateway/GatewayRouterFacet.sol
aakoshh: We need the validator index as input because it's required for checking the proof. Basically what the index provides when converted to binary form is the instruction whether to go left or right in the tree to reach the leaf in which the value is supposed to be stored, which guides the algorithm in choosing which order to apply hashes in.

I explained it [here](https://github.com/aakoshh/metronome/blob/e5b657b0ac68fe3bf37ee0c11fb8f6f55ea0e7ec/metronome/checkpointing/models/src/io/iohk/metronome/checkpointing/models/MerkleTree.scala#L27-L70) once.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -30,6 +33,9 @@ contract GatewayRouterFacet is GatewayActorModifiers {
     using FvmAddressHelper for FvmAddress;
     using StorableMsgHelper for StorableMsg;
 
+    event QuorumReached(uint64 height, bytes32 checkpoint, Membership membership, uint256 weight);
+    event QuorumUpdated(uint64 height, bytes32 checkpoint, Membership membership, uint256 weight);
```
aakoshh, line 37: To be honest I wouldn't put the entire `Membership` in the event. It can be retrieved from CometBFT. If anything add the key that signed it. 

It's also not clear from looking at this whether `weight` was the increase, or the total weight.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +185,59 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        uint64 validatorIndex,
+        bytes calldata signature
+    ) external systemActorOnly {
+        (address recoveredSignatory, ECDSA.RecoverError err, ) = ECDSA.tryRecover(checkpoint, signature);
+        if (err != ECDSA.RecoverError.NoError) {
+            revert InvalidSignature();
+        }
+
+        if (s.bottomUpCollectedSignatures[height][signature]) {
+            revert RepliedSignature();
```
aakoshh, line 208:
```suggestion
            revert DuplicateSignature();
```
It's up to you but I didn't know what "replied" means, it sounded like "we already replied". The signature isn't really a reply to any request per se.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +185,59 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        }
+
+        if (validatorIndex >= membership.validators.length) {
+            revert InvalidValidatorIndex();
+        }
+
+        address validator = membership.validators[validatorIndex].addr.extractEvmAddress().normalize();
+
+        if (validator == recoveredSignatory) {
+            s.bottomUpCollectedSignatures[height][signature] = true;
```
aakoshh, line 218: Wait, I don't think signatures are unique, they have some random element in them. If you use something for indexing it should be the address.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +185,59 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+
+            CheckpointQuorum memory quorum = s.bottomUpCheckpointQuorum[height];
+            quorum.weight += membership.validators[validatorIndex].weight;
+
+            if (quorum.weight >= membership.totalWeight) {
```
aakoshh, line 223: This seems to require 100% of the signatures to be present, rather than 2/3.

aakoshh: There is a [majority_percentage](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/genesis/src/lib.rs#L140) field in genesis which gets [passed](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/actor_interface/src/ipc.rs#L95) to the gateway. Shouldn't we use that here?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +185,59 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+    /// @param validatorIndex - the index of the validator in the membership that signed the checkpoint
+    /// @param signature - the signature of the checkpoint
+    function addCheckpointSignature(
+        uint64 height,
+        bytes32 checkpoint,
+        Membership calldata membership,
+        uint64 validatorIndex,
+        bytes calldata signature
+    ) external systemActorOnly {
+        (address recoveredSignatory, ECDSA.RecoverError err, ) = ECDSA.tryRecover(checkpoint, signature);
```
aakoshh, line 202: In this context `checkpoint` should not be an input. It should have been stored in https://github.com/consensus-shipyard/fendermint/issues/248 and now retrieved by height.

src/gateway/GatewayRouterFacet.sol
cryptoAtwill: Oh, I see. If this is the only reason, I think we dont need this for the contract. This is the example provided by openzepplin: https://github.com/OpenZeppelin/merkle-tree. As long as offchain knows the index, or can derive the index, it should be fine. At least for the openzeppline library, there is no need to attach.

aakoshh: It would be nice to understand how that works. Are they encoding the left/right hints into the proof somehow? Is there a Rust implementation of this OpenZeppelin algorithm?

aakoshh: Checking https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MerkleProof.sol

aakoshh: It looks like the secret is in the [sorted property](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MerkleProof.sol#L215).

aakoshh: For the record https://github.com/literallymarvellous/merkle-tree-rs is purpose made to work with the above Solidity and Javascript library.

aakoshh: Added the conclusion to https://github.com/consensus-shipyard/fendermint/issues/249#issuecomment-1725275034

I reckon the method signature is:

```
addSignature(uint64 height, bytes32[] proof, u256 weight, bytes signature)
```

src/gateway/GatewayRouterFacet.sol
dnkolegov: > but in solidity, it's U256
@cryptoAtwill In the Solidity actors we also use `uint64` as far as I see

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: That was going to be `Replayed`

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,69 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+    /// @param height - the height of the block in the checkpoint
+    /// @param proof - a Merkle proof that the validator was in the membership at height `h`
+    /// @param weight - the weight of the validator
+    /// @param signature - the signature of the checkpoint
+    function addCheckpointSignature(
+        uint64 height,
+        bytes32[] memory proof,
+        uint256 weight,
+        bytes memory signature
+    ) external systemActorOnly {
```
aakoshh, line 200: This can now be called by non-system actors.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,69 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        BottomUpCheckpointNew memory checkpoint = s.bottomUpCheckpoints[height];
+
+        CheckpointMembership memory membership = s.bottomUpCheckpointMembership[height];
+
+        // TODO: do we need a helper here?
+        bytes32 checkpointHash = keccak256(abi.encode(checkpoint));
```
aakoshh, line 206: You could store this in the `CheckpointMembership` so it doesn't need to be recalculated for each signature.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -24,16 +24,24 @@ struct GatewayActorStorage {
...
+    /// @notice A mapping of block numbers to bottom-up checkpoints
+    // slither-disable-next-line uninitialized-state
+    mapping(uint64 => BottomUpCheckpointNew) bottomUpCheckpoints;
+    /// @notice A mapping of block numbers to bottom-up checkpoint membership data
+    // slither-disable-next-line uninitialized-state
+    mapping(uint64 => CheckpointMembership) bottomUpCheckpointMembership;
+    /// @notice The validators have already sent signatures at height `h`
+    mapping(uint64 => mapping(address => bool)) bottomUpCollectedSignatures;
+    /// @notice The signatures collected for the checkpoint at height `h`
+    mapping(uint64 => CheckpointQuorum) bottomUpCheckpointQuorum;
```
aakoshh, line 44: I'm curious if you have any expectation of gas costs based on whether `CheckpointMembership` and `CheckpointQuorum` are kept as separate records in separate maps, or whether they could be merged into one?

aakoshh (commented):
This looks good now, just minor comments and questions :+1:

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +186,69 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        if (!valid) {
+            revert FailedToVerifyValidatorMembership();
+        }
+
+        s.bottomUpCollectedSignatures[height][recoveredSignatory] = true;
+
+        CheckpointQuorum memory quorum = s.bottomUpCheckpointQuorum[height];
+        quorum.weight += weight;
+
+        if (quorum.weight >= membership.totalWeight) {
```
aakoshh, line 229: Hm, this still doesn't take the `majority_percentage` into account.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/structs/Checkpoint.sol
```diff
@@ -21,6 +21,26 @@ struct BottomUpCheckpoint {
     bytes proof;
 }
 
+/// @notice A bottom-up checkpoint struct.
+// TODO: Remove old BottomUpCheckpoint and update all the codebase.
+struct BottomUpCheckpointNew {
+    /// @dev Child subnet ID, for replay protection from other subnets where the exact same validators operate.
+    /// Alternatively it can be appended to the hash before signing, similar to how we use the chain ID.
+    SubnetID subnetID;
```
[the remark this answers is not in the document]
dnkolegov: I did that because we used it even in previous `BottomUpCheckpoint`.

src/lib/LibGatewayActorStorage.sol
dnkolegov: I did that not to save gas but just because, at that time, we did not have all these new data types. Now, they can be merged.

src/gateway/GatewayRouterFacet.sol
dnkolegov: I added it before it was used in the Voting Library only. Please check. Thank you.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Now fixed.

dnkolegov:
> This looks good now, just minor comments and questions 👍

Great catches!
Thank you @aakoshh !

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +187,101 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        if (checkpoint.blockHeight == 0) {
+            revert CheckpointNotCreated();
+        }
+
+        CheckpointInfo memory checkpointInfo = s.bottomUpCheckpointInfo[height];
+        if (checkpointInfo.rootHash.length == 0) {
+            revert CheckpointMembershipNotCreated();
+        }
+
+        bytes32 checkpointHash = checkpoint.toHash();
```
cryptoAtwill, line 212: is there a difference btw `checkpointHash` and `checkpointInfo.hash`?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -179,4 +187,101 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+    /// @param membershipProof - a Merkle proof that the validator was in the membership at height `h`
+    /// @param weight - the weight of the validator
+    /// @param signature - the signature of the checkpoint
+    function addCheckpointSignature(
+        uint64 height,
+        bytes32[] memory membershipProof,
+        uint256 weight,
+        bytes memory signature
+    ) external {
+        BottomUpCheckpointNew memory checkpoint = s.bottomUpCheckpoints[height];
```
cryptoAtwill, line 202: do we do backward check? Like the height should not have already committed?

src/gateway/GatewayRouterFacet.sol
cryptoAtwill: If the `checkpoint` is read from storage, that means we can storage the `checkpointHash` in storage as well? So that we dont have to compute it every time.

dnkolegov: I forgot checnge that one. Thanks!

src/gateway/GatewayRouterFacet.sol
dnkolegov: The current idea is to continue to allow accumulating signatures even if it has been committed. In this case, we update signatures and emit an event with updated weight. If I understood you right...

src/gateway/GatewayRouterFacet.sol
aakoshh: I'm sorry what was the fix? I'm not a Solidity storage buff; it looks like you can take records and modify them and expect them to be persisted, I'm just checking if that's the case.

aakoshh (approved):
Bullseye :dart:

src/gateway/GatewayRouterFacet.sol
dnkolegov: The fix is to use "storage" type instead of "memory" type. But it was not fixed actually in that sense. Fixed now. Thank you.

src/structs/Checkpoint.sol
```diff
@@ -21,6 +21,39 @@ struct BottomUpCheckpoint {
...
 
+/// @notice A bottom-up checkpoint type.
+// TODO: Remove old BottomUpCheckpoint, rename BottomUpCheckpointNew to BottomUpCheckpoint and update the codebase.
+struct BottomUpCheckpointNew {
+    /// @dev Child subnet ID, for replay protection from other subnets where the exact same validators operate.
+    /// Alternatively it can be appended to the hash before signing, similar to how we use the chain ID.
+    SubnetID subnetID;
+    /// @dev The height of the child subnet at which this checkpoint was cut.
+    /// Has to follow the previous checkpoint by checkpoint period.
+    uint64 blockHeight;
```
aakoshh, line 32: Maybe we can add the block hash to it as well? Just so it also acts as an actual checkpoint?

dnkolegov: Added - https://github.com/consensus-shipyard/ipc-solidity-actors/pull/200/commits/a783e1263b8e35c093f7db273c9adf694f69c073

src/structs/Checkpoint.sol
```diff
@@ -21,6 +21,39 @@ struct BottomUpCheckpoint {
     bytes proof;
 }
 
+/// @notice A bottom-up checkpoint type.
+// TODO: Remove old BottomUpCheckpoint, rename BottomUpCheckpointNew to BottomUpCheckpoint and update the codebase.
```
adlrocha, line 25: @dnkolegov, @cryptoAtwill, are we tracking this TODO somewhere?

src/gateway/GatewayGetterFacet.sol
```diff
@@ -50,8 +50,9 @@ contract GatewayGetterFacet {
         return s.networkName;
     }
 
+    // TODO: remove or add a new getter
```
adlrocha, line 53: @dnkolegov, should we track this in an issue like the others TODO?

dnkolegov: https://github.com/consensus-shipyard/ipc-solidity-actors/issues/210

Issue #203, opened 2023-09-18
Update `join` and `leave` subnet


cryptoAtwill:
### Motivation
As we are relying on `fendermint` to perform top down and bottom up checkpointing, the existing way to update validator set and collateral management might have to be updated. 
Checkpoints are submitted at a fixed interval and validator sets are updated at the fixed interval. If the validator leaves the subnet and withdraws the collateral immediately, one cannot guarantee the validator would behave legitimately. We should not allow validator to withdraw immediately and only withdraw the collateral after the next checkpoint is committed in the child chain.

#### Parent Chain
When `join` and `leave` is called, the changes are committed to chain (we can also just emit an event, but based on the current util infra, committing to chain is faster to implement). For example, we can have `ValidatorChangeSet` struct in the subnet with something like the below:
```solidity
struct ValidatorChangeSet {
    /// WIth each update, the configuration number increments by 1
    configuration_number: u64,
    operation: Join/Leave,
    worker_address: FvmAddress,
    address: address,
    /// Some extra data payload the callers should interpret themselves
    extra_payload: bytes
}
``` 
Then we could have an array that keeps the changes:
```solidity
validatorChanges: ValidatorChangeSet[]
```
Then `join` and `leave` are just simply appending the `validatorChanges`.

#### Top Down Finality
We add a getter function `getValidatorChanges` to get the unexecuted validator changes, this will be used by top down parent finality.
We add an `applyValidatorChanges` function in the `GatewayManagerFacet` that applies the changes, like how it's done currently. In `fendermint` top down parent finality, after the proposals are finalised, it will call `applyValidatorChanges`. This relays the validator set changes to the child chain.

#### Bottom Up Finality
In bottom up checkpoint, the validators submits the checkpoint at a fixed interval. For the next submission epoch, validators need to query the validator set in the child chain from the `SubnetActor` contract. Validators will obtain the last executed configuration number for that epoch and its corresponding validator set, we name this set as `OldValidatorSet`. The signatures for the bottom up checkpoint must be collected from this group of `OldValidatorSet` and propagate to the parent chain. 
_**(Question1: what if we cannot get all the signatures from all the validators in `OldValidatorSet`? Will it block the bottom up checkpoint in anyway? Or we just need from some of the validators?)**_

We also need to add one more field to the parent gateway
```solidity
childAppliedConfigurationNumber: u64
```
[...]

PR #204, opened 2023-09-19
Update join and leave

cryptoAtwill:
This PR updates the `join` and `leave` function call in the subnet manager. So that we can track the changes instead for the checkpointing to apply the changes.

It is implemented as described in: https://github.com/consensus-shipyard/ipc-solidity-actors/issues/203

### DIFF
--- a/src/lib/LibSubnetActorStorage.sol
+++ b/src/lib/LibSubnetActorStorage.sol
@@ -56,11 +56,11 @@ struct SubnetActorStorage {
     // ============ Validator Related ===========
     /// @notice Sequence number that uniquely identifies a validator set.
     uint64 configurationNumber;
-    /// @notice The set of validators changes to be committed
+    /// @notice The set of validators changes to be committed.
     ValidatorChange[] validatorSetChanges;
-    /// @notice List of validators in the subnet that are currently
+    /// @notice List of validators in the subnet that are currently active.
     EnumerableSet.AddressSet validators;
-    /// @notice address to validator information
+    /// @notice address to validator information.
     mapping(address => Validator) validatorInfo;
 }
 
--- a/src/lib/LibSubnetActorStorage.sol
+++ b/src/lib/LibSubnetActorStorage.sol
@@ -7,7 +7,7 @@ import {BottomUpCheckpoint} from "../structs/Checkpoint.sol";
 import {NotGateway, SubnetAlreadyKilled} from "../errors/IPCErrors.sol";
 import {EpochVoteBottomUpSubmission} from "../structs/EpochVoteSubmission.sol";
 import {FvmAddress} from "../structs/FvmAddress.sol";
-import {SubnetID} from "../structs/Subnet.sol";
+import {SubnetID, Validator, ValidatorChange} from "../structs/Subnet.sol";
 import {Address} from "openzeppelin-contracts/utils/Address.sol";
 import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
 
@@ -31,8 +31,6 @@ struct SubnetActorStorage {
     uint256 totalStake;
     /// @notice Minimal activation collateral
     uint256 minActivationCollateral;
-    /// @notice Sequence number that uniquely identifies a validator set.
-    uint64 configurationNumber;
     /// @notice number of blocks in a top-down epoch
     uint64 topDownCheckPeriod;
     /// @notice number of blocks in a bottom-up epoch
@@ -50,12 +48,20 @@ struct SubnetActorStorage {
     /// @notice Type of consensus algorithm.
     /// @notice current status of the subnet
     Status status;
-    /// @notice List of validators in the subnet
-    EnumerableSet.AddressSet validators;
     /// @notice ID of the parent subnet
     SubnetID parentId;
     /// immutable params
     ConsensusType consensus;
+
+    // ============ Validator Related ===========
+    /// @notice Sequence number that uniquely identifies a validator set.
+    uint64 configurationNumber;
+    /// @notice The set of validators changes to be committed
+    ValidatorChange[] validatorSetChanges;
+    /// @notice List of validators in the subnet that are currently
+    EnumerableSet.AddressSet validators;
+    /// @notice address to validator information
+    mapping(address => Validator) validatorInfo;
 }
 
 library LibSubnetActorStorage {
--- a/src/structs/Subnet.sol
+++ b/src/structs/Subnet.sol
@@ -13,6 +13,30 @@ struct SubnetID {
     address[] route;
 }
 
+/// @notice All the information about a validator
+struct Validator {
+    address validator;
+    uint256 weight;
+    /// @notice The extra info about the validator. This information is not
+    /// @notice important for the contract, but useful for offchain computation.
+    /// @dev Offchain should know how to decode these bytes, contract does not
+    /// @dev need to know about the details.
+    bytes info;
+}
+
+/// @notice All the information about a validator
+struct ValidatorChange {
+    Validator validator;
+    uint64 configurationNumber;
+    SubnetOpt operation;
+}
+
+/// The operations that validators can perform for the subnet lifecycle
+enum SubnetOpt {
+    Join,
+    Leave
+}
+
 struct Subnet {
     uint256 stake;
     uint256 genesisEpoch;
--- a/src/subnet/SubnetActorManagerFacet.sol
+++ b/src/subnet/SubnetActorManagerFacet.sol
@@ -258,7 +258,7 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         uint256 length = s.validatorSetChanges.length;
 
         if (length != 0) {
-            // ensure the configuration number is consequential
+            // ensure the configuration number is sequential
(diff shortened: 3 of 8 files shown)


src/lib/LibSubnetActorStorage.sol
```diff
@@ -50,12 +48,20 @@ struct SubnetActorStorage {
...
     /// @notice ID of the parent subnet
     SubnetID parentId;
     /// immutable params
     ConsensusType consensus;
+
+    // ============ Validator Related ===========
+    /// @notice Sequence number that uniquely identifies a validator set.
+    uint64 configurationNumber;
+    /// @notice The set of validators changes to be committed
+    ValidatorChange[] validatorSetChanges;
```
dnkolegov, line 60: Is it possible to remove validators from this array when necessary?

src/lib/LibSubnetActorStorage.sol
```diff
@@ -50,12 +48,20 @@ struct SubnetActorStorage {
...
+    /// @notice List of validators in the subnet that are currently
```
dnkolegov, line 61: "that are currently" ?

cryptoAtwill: oops

src/lib/LibSubnetActorStorage.sol
cryptoAtwill: right, when bottom up is committed, we should have them removed.

dnkolegov: Adding a todo or opening an issue would be helpful to not forget about it.

src/lib/LibSubnetActorStorage.sol
```diff
@@ -50,12 +48,20 @@ struct SubnetActorStorage {
...
-    /// @notice List of validators in the subnet
-    EnumerableSet.AddressSet validators;
     /// @notice ID of the parent subnet
     SubnetID parentId;
     /// immutable params
     ConsensusType consensus;
+
+    // ============ Validator Related ===========
+    /// @notice Sequence number that uniquely identifies a validator set.
+    uint64 configurationNumber;
```
aakoshh, line 58: This comment doesn't really tell me what this number is:
* is it the ID of the last known committed configuration in the subnet?
* is it the ID used for the next change to be enqueued in `validatorSetChanges`?

We can use docstrings to explain it as if it was for a person who has never heard anything about this before.

src/lib/LibSubnetActorStorage.sol
aakoshh: This is another one where more elaborate docstrings can help explain that these are queued changes not yet confirmed by the subnet validators, waiting to be rolled into the current validators up to the configuration number we see in the next bottom-up checkpoint.

src/lib/LibSubnetActorStorage.sol
```diff
@@ -50,12 +48,20 @@ struct SubnetActorStorage {
...
+    /// @notice The set of validators changes to be committed.
+    ValidatorChange[] validatorSetChanges;
+    /// @notice List of validators in the subnet that are currently active.
+    EnumerableSet.AddressSet validators;
+    /// @notice address to validator information.
+    mapping(address => Validator) validatorInfo;
```
aakoshh, line 64: Does this contain the _current_ validators with their _current_ weight?

src/structs/Subnet.sol
```diff
@@ -13,6 +13,30 @@ struct SubnetID {
...
+    address validator;
+    uint256 weight;
+    /// @notice The extra info about the validator. This information is not
+    /// @notice important for the contract, but useful for offchain computation.
+    /// @dev Offchain should know how to decode these bytes, contract does not
+    /// @dev need to know about the details.
+    bytes info;
+}
+
+/// @notice All the information about a validator
```
aakoshh, line 27: Probably copy-pasted comment.

src/structs/Subnet.sol
```diff
@@ -13,6 +13,30 @@ struct SubnetID {
...
+struct ValidatorChange {
+    Validator validator;
+    uint64 configurationNumber;
+    SubnetOpt operation;
+}
+
+/// The operations that validators can perform for the subnet lifecycle
+enum SubnetOpt {
```
aakoshh, line 35: "Opt" usually stands for "Option", and "Op" is for "Operation"

src/structs/Subnet.sol
```diff
@@ -13,6 +13,30 @@ struct SubnetID {
...
+    /// @dev Offchain should know how to decode these bytes, contract does not
+    /// @dev need to know about the details.
+    bytes info;
+}
+
+/// @notice All the information about a validator
+struct ValidatorChange {
+    Validator validator;
+    uint64 configurationNumber;
+    SubnetOpt operation;
```
aakoshh, line 31: Do we really need this? 
* Once you join, if you raise or lower your collateral, are you joining again with some new weight?
* Can you join with 0 weight?
* When you leave, does your weight matter?

The change of a validator to a new weight seems sufficient to express what's going on, with weight = 0 meaning leaving the active validator set.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -31,8 +31,60 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     event BottomUpCheckpointSubmitted(BottomUpCheckpoint checkpoint, address submitter);
     event BottomUpCheckpointExecuted(uint64 epoch, address submitter);
     event NextBottomUpCheckpointExecuted(uint64 epoch, address submitter);
+    /// @notice Emitted when there is an update in the validator set.
+    event ValidatorUpdate(ValidatorChange change);
+
+    /// @notice method that allows a potential validator wants to join the subnet
+    /// @param extraInfo the extra data associated with this join operation.
+    function joinSubnet(bytes calldata extraInfo) external payable notKilled {
+        uint256 validatorStake = msg.value;
```
aakoshh, line 40: What happens if you join multiple times? Doesn't your stake add up?

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -200,4 +252,21 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
 
         IGateway(s.ipcGatewayAddr).commitChildCheck(checkpoint);
     }
+
+    /// @notice append a new validator change, ensures the configuration number in the updates are sequential
+    function _appendValidatorChange(ValidatorChange memory change) internal {
+        uint256 length = s.validatorSetChanges.length;
+
+        if (length != 0) {
+            // ensure the configuration number is consequential
```
aakoshh, line 261:
```suggestion
            // ensure the configuration number is sequential
```

aakoshh (commented):
Thank you for taking the initiative on this issue! 

I know very little about what you did for staking, so I'm just going to ask some random things and see what sticks:

* It looks like the idea is that a validator joins with some collateral that gives them weight, they go in the queue, at some point become confirmed. They have to be confirmed if they want to leave, and if they are then their weight from when they were confirmed goes again into the queue and if their leaving gets confirmed they can get their collateral back. 
* Currently there is no handling of adjusting the amount of collateral. You can actually join twice, and perhaps it would be possible for rolling up multiple joins into a grand total weight (we need to specify if they are considered absolute values or deltas), but when you leave, you are assumed to want to take it down to zero. 
* Because the contract allows joining and leaving multiple times, it is possible to end up with behaviour that I find hard to reason about, e.g. `join 1, join 2, commit 1, leave 1, leave 1, leave 1, ...`. 
* The contract doesn't seem to consider the limits to how many validators can join. For example CometBFT would work with up to 100 validators, so at any point in time the contract should send to the subnet the top 100 validators who are eligible to produce votes, with their power based on their collateral, and that could mean that some validators actually leave the pool because others have joined. However here we just send them on regardless of how many are already in the queue. Do you expect the subnet to do the calculation? 
* If the answer to the previous question is "Yes" then I'd like to point out that not everybody in a configuration number is eligible to produce blocks and sign checkpoints, only some unknown subset of validators who made the cut. It would be better IMO if the SubnetActor had a proper staking module. 
* Not to even mention delegated staking, which I think is what anyone would assume these days.
* If the answer is "No" then I think a configuration number should cover more than a single validator change; rather it's the set of changes triggered by the action of some validator.
* For the future: when a validator leaves the subnet, they have to go into an unbond queue before they can get their collateral back, so that evidence of fraud can be raised against them for some time.


An alternative design could looks something like this:
* There is a single staking distribution table. 
* There is a _current_ validator set and a _next_ validator set derived from the staking distribution, based on some rules such as taking the top 100 validators based on (delegated) stake.
* When a validator joins, or is delegated more collateral, their stake is updated in the staking distribution.
* When the staking distribution is updated, it causes a recalculation of the _next_ validator set. The _changes_ made to the validator set in this step become a single atomic configuration number. For example if a validator joins with collateral higher than the 100th validator in the table, then the changeset will be that the new joiner gets their power, while the one that fell off the cliff gets 0 power. Or if some validators are slashed, it could mean multiple validators entering the active set.
* It is possible that power does not equal collateral.
* The changeset is immediately applied on the _next_ validator set and sent to the queue. 
* Confirmed changes are applied on the _current_ validator set. 
* Leaving or otherwise adjusting the collateral is based on the single staking distribution table, not the current validator set, so you always know how much a person can try to take back.

PR #205, opened 2023-09-20
Rust binding

cryptoAtwill:
Shifting rust binding into this repo so that `ipc-agent` and `fendermint` can import the bindings instead. Changes:
- Adding `make rust-binding` to build the binding, taken the code from existing `fendermint` with some slight changes.
- Adding `husky` to enforce a binding check, ensures the binding is updated with new push.
- Perform type conversion from fvm address to FvmAddress.

One can just import in rust with:
```
ipc_actors_abis = { git = "https://github.com/consensus-shipyard/ipc-solidity-actors.git", branch = "dev" }
```

One draw back with this approach is whoever imports this `ipc_actors_abis` also downloads the submodule in the solidity contracts, like openzeppline.


binding/build.rs
```diff
@@ -0,0 +1,115 @@
...
+/// so we have better code completion with Rust Analyzer.
+fn main() {
+    // Run with `cargo build -vv` to see output from any `eprintln!` or `println!`.
+
+    // We are not building anything, could be imported as crate.
+    if std::env::var("BUILD_BINDINGS").ok().is_none() {
+        return;
+    }
+
+    // Use the env var set by the Makefile, or fall back to the default.
```
aakoshh, line 19: This comment is no longer accurate, there is no `IPC_ACTORS_DIR` env var like in the fendermint repo.

binding/src/convert.rs
```diff
@@ -0,0 +1,100 @@
...
+                Ok(addr)
+            }
+        }
+    };
+}
+
+/// Converts a Rust type FVM address into its underlying payload
+/// so it can be represented internally in a Solidity contract.
+pub fn addr_payload_to_bytes(
+    payload: fvm_shared::address::Payload,
```
aakoshh, line 33: I would define the two functions as `addr_to_bytes` and `bytes_to_addr` and work on `fvm_shared::address::Address`, to make it symmetrical.

binding/src/convert.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/// It takes the bytes from an FVMAddress represented in Solidity and
+/// converts it into the corresponding FVM address Rust type.
+pub fn bytes_to_fvm_addr(
+    protocol: u8,
+    bytes: &[u8],
+) -> anyhow::Result<fvm_shared::address::Address> {
+    let addr = match protocol {
+        1 => fvm_shared::address::Address::from_bytes(&[[1u8].as_slice(), bytes].concat())?,
```
aakoshh, line 57: This is difficult to read, I have been staring at tit for minutes thinking you are passing two distinct arguments here.

binding/src/convert.rs
aakoshh: Maybe they can work with a tuple of `(u8, Bytes)`, or be made private. They are not the most intuitive to use on their own.

binding/src/convert.rs
```diff
@@ -0,0 +1,100 @@
...
+    match payload {
+        fvm_shared::address::Payload::Secp256k1(v) => Ok(ethers::types::Bytes::from(v)),
+        fvm_shared::address::Payload::Delegated(d) => {
+            let addr = d.subaddress();
+            let b = ethers::abi::encode(&[ethers::abi::Token::Tuple(vec![
+                ethers::abi::Token::Uint(ethers::types::U256::from(d.namespace())),
+                ethers::abi::Token::Uint(ethers::types::U256::from(addr.len())),
+                ethers::abi::Token::Bytes(addr.to_vec()),
+            ])]);
+            Ok(ethers::types::Bytes::from(b))
```
aakoshh, line 44: Why do you need this complicated ABI dance instead of just using `Payload::to_raw_bytes` and `Payload::new`? Why do we have to pack the bytes differently from how it's already done in `fvm_shared`?

aakoshh (commented):
Nice one! Did you want to add the internal type conversions later?

cryptoAtwill:
@aakoshh I added those type conversion in `ipc-sdk`, that's where all the other types are defined, which should save a lot of type conversion in `fendermint`.

binding/src/convert.rs
cryptoAtwill: I did one round of checking, seems like in `ipc-agent`, these functions are not needed. I will just make them private and they are util functions for type conversion.

binding/src/convert.rs
cryptoAtwill: Yeah, good one. Perhaps this is just a short cut but the reason for this was in solidity we have to recover these bytes into `DelegatedAddress`, I'm not sure how it was actually done in fvm library and not sure if it's easy to recover in solidity, hence this implementation that is easy to parse and create in solidity.

aakoshh:
How can you add type conversions for types defined in these modules in a different crate? Like how can you do a `impl From<gateway_manager_facet::SubnetID> for gateway_getter_facet::SubnetID` in a different crate?

.husky/pre-push
```diff
@@ -0,0 +1,23 @@
...
+# sometime one cannot compile abi on mac, just a way to ignore compilation
+# if one really cannot compile abi and obtains the `out` folder from other
+# sources. Use with care.
+if [[ ! $SKIP_COMPILE ]]; then
+    make compile-abi
+fi
+
+# update the rust binding if we need it
+if [[ ! $SKIP_RUST_BINDING ]]; then
+    make rust-binding
```
aakoshh, line 13: So this should prevent anyone from pushing invalid code to the repo, right? It would still be nice if CI did the build as well, just to see that it compiles for sure.

cryptoAtwill:
@aakoshh oh. I misunderstood you, I thought you are saying converting `ipc_sdk::SubnetID` to `gateway_getter_facet::SubnetID`. 
Yeah, this kind of conversion needs to be done here. I only added the ones I know will be used, I can add `SubnetID` to the facets as well.

.husky/pre-push
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
aakoshh, line 2: I don't have Husky. Where does it say hot to install it?

.husky/pre-push
cryptoAtwill: right, initial intention is just for those who don't really need rust binding, like updating readme, can save the troubles from building rust crates.

.husky/pre-push
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env sh
```
aakoshh, line 1: Should this be `bash`? 

On my Ubuntu machine I tried the following:

```console
$ cat test.sh
#!/usr/bin/env sh

if [[ ! $SKIP_RUST_BINDING ]]; then
  echo "building..."
fi
$ ./test.sh
./test.sh: 3: [[: not found

```

It works with `bash`.

.husky/pre-push
cryptoAtwill: it should be auto installed when you do `npm install --dev`.

.husky/pre-push
aakoshh: Yeah, I'm not arguing against the `SKIP_RUST_BINDING`, just wondering if there should be some sanity check on CI. I guess it's okay if not.

binding/src/convert.rs
aakoshh: For the record it's here: https://github.com/filecoin-project/ref-fvm/blob/fvm%40v3.8.0/shared/src/address/payload.rs#L115

binding/build.rs
```diff
@@ -0,0 +1,109 @@
...
+    let ipc_actors_dir = workspace_dir()
+        .parent()
+        .unwrap()
+        .to_string_lossy()
+        .into_owned();
+
+    let lib_path = format!("{ipc_actors_dir}/binding/src/lib.rs");
+    let mut lib = std::fs::File::create(lib_path).expect("failed to create lib.rs");
+    writeln!(lib, "// DO NOT EDIT! This file was generated by build.rs").unwrap();
+    writeln!(lib, "#[macro_use]\npub mod convert;").unwrap();
```
aakoshh, line 28: Probably doesn't need to be `pub` any more.

aakoshh (approved):
LGTM! 

I wonder if there is any way to find where cargo downloads the source so we can get the actual Solidity files from it as well for the Docker.

adlrocha (approved):
Thank you for this one, Will. I am reviewing from my phone so I may have left something behind, but overall LGTM. It will make our lives easier :) 

Let's propagate this one to the agent and Fendermint.

PR #206, opened 2023-09-21
Add rust binding to CI

cryptoAtwill:
Adding rust binding check to CI.


adlrocha (approved):
LGTM, but I am just wondering why we may be getting different bindings according to the runtime. That is weird.

PR #207, opened 2023-09-21
Method to list incomplete checkpoints

dnkolegov:
This PR adds a method to list incomplete checkpoints described in https://github.com/consensus-shipyard/fendermint/issues/258 and a method to garbage collect completed checkpoints


src/gateway/GatewayRouterFacet.sol
```diff
@@ -199,6 +202,9 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         uint256 weight,
         bytes memory signature
     ) external {
+        if (height < s.bottomUpCheckpointRetentionIndex) {
+            revert CheckpointAlreadyProcess();
```
adlrocha, line 206: typo?
```suggestion
            revert CheckpointAlreadyProcessed();
```

src/gateway/GatewayRouterFacet.sol
```diff
@@ -285,4 +302,30 @@ contract GatewayRouterFacet is GatewayActorModifiers {
             reached: false
         });
     }
+
+    /// @notice garbage collect all checkpoints and related data for heights lower than `newIndex`.
+    /// @param newIndex - new retention index
+    function pruneBottomUpCheckpoints(uint64 newIndex) external systemActorOnly {
```
adlrocha, line 308: When is this function expected to be called externally? The idea is that this function should be called when there is no longer need to store the checkpoint signatures because it has already been submitted up, right? 

If this is the case, would you mind adding this description in the comments so we are explcit?

adlrocha, line 308: How do we expect to set the `newIndex` for a specific garbage collection?

src/lib/LibGatewayActorStorage.sol
```diff
@@ -38,8 +39,15 @@ struct GatewayActorStorage {
     /// @notice A mapping of block numbers to checkpoint data
     // slither-disable-next-line uninitialized-state
     mapping(uint64 => CheckpointInfo) bottomUpCheckpointInfo;
+    /// @notice The height of the last bottom-up checkpoint registered in the parent.
+    /// All checkpoint with the height less than or equal to that number can be garbage collected in the child subnet.
+    /// @dev Initial retention index is 1.
+    uint64 bottomUpCheckpointRetentionIndex;
```
adlrocha, line 45: Do you mind explaining here that the retention is needed so the checkpoint and signatures are available so relayers can submit the chceckpoint? It wasn't clear to me the purpose of the `RetentionIndex` on a first read. Thanks!

adlrocha (approved):
A few cosmetic changes to make the code more understandable but overall LGTM.  Thanks. I am curious how the end-to-end will work, and the gas cost from this (but it probably doesn't apply for now).

src/lib/LibGatewayActorStorage.sol
dnkolegov: So 

- https://github.com/consensus-shipyard/fendermint/issues/258 wants to retrieve the checkpoints that have not reached quorum yet
- https://github.com/consensus-shipyard/fendermint/issues/250 wants to clear the checkpoints registered (applied) in the parent
-
`retentionIndex` is the number of the first incomplete checkpoints we must have to implement checkpointing.

All checkpoints with a height less than `retentionIndex` may be removed from the history if we want.

src/gateway/GatewayRouterFacet.sol
dnkolegov: If I understand correctly, the caller should know what checkpoints have been processed in the parent and can be removed in the child

dnkolegov: > The idea is that this function should be called when there is no longer need to store the checkpoint signatures because it has already been submitted up, right

correct

dnkolegov: Updated description

src/lib/LibGatewayActorStorage.sol
adlrocha: I meant in the code, so it is clear :) But no need if you feel it may be redundant or clear once the end-to-end is implemented.

adlrocha (commented):
Unless @aakoshh  and @cryptoAtwill want to have a final look, I think is ready to :ship:

src/lib/LibGatewayActorStorage.sol
dnkolegov: Updated the code

src/gateway/GatewayGetterFacet.sol
```diff
@@ -194,4 +196,34 @@ contract GatewayGetterFacet {
...
+        return checkpoints;
+    }
+
+    /// @notice get the bottom-up checkpoint retention index
+    function getBottomUpRetentionIndex() public view returns (uint64) {
+        return s.bottomUpCheckpointRetentionIndex;
+    }
+
+    /// @notice get the threshold value for child subnet
+    function getChildThreshold(uint256 weight) public view returns (uint256) {
```
aakoshh, line 226: Just by reading this I have a hard time understanding what it's about. Perhaps it could be explained more in the docstring. 

Don't all children have their own thresholds? What is `weight` and what does it return? What can we use it for? Which child?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -233,12 +240,20 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         // All checks passed.
         // Adding signature and emitting events.
 
-        s.bottomUpCollectedSignatures[height][recoveredSignatory] = true;
+        bool ok = s.bottomUpCollectedSignatures[height].add(recoveredSignatory);
+        if (!ok) {
+            revert FailedAddIncompleteCheckpoint();
```
aakoshh, line 245: I don't understand what this error means here. We tried to add a signature that we already checked wasn't a replay, but failed. How can it fail, and why does it say "failed to add an incomplete checkpoint"?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -233,12 +240,20 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
         checkpointInfo.currentWeight += weight;
 
         if (checkpointInfo.currentWeight >= checkpointInfo.threshold) {
             if (!checkpointInfo.reached) {
                 checkpointInfo.reached = true;
+                // checkpoint is completed since the threshold has been reached
+                ok = s.incompleteCheckpoints.remove(checkpoint.blockHeight);
+                if (!ok) {
+                    revert FailedRemoveIncompleteCheckpoint();
+                }
```
aakoshh, line 256: How can this fail? Wouldn't that be a programming error in the contract, or is it somehow the caller's fault?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -263,20 +278,28 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
 
         uint256 threshold = LibGateway.getThreshold(membershipWeight);
 
+        // process the checkpoint
         s.bottomUpCheckpoints[checkpoint.blockHeight] = checkpoint;
-
+        bool ok = s.incompleteCheckpoints.add(checkpoint.blockHeight);
+        if (!ok) {
+            revert FailedAddIncompleteCheckpoint();
+        }
```
aakoshh, line 302: Why would this fail?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -285,4 +308,28 @@ contract GatewayRouterFacet is GatewayActorModifiers {
             reached: false
         });
     }
+
+    /// @notice set a new retention index and garbage collect all checkpoints in range [`retentionIndex`, `newRetentionIndex`)
+    /// @dev `retentionIndex` is the number of the first incomplete checkpoints we must have to implement checkpointing
+    /// All checkpoints with a height less than `retentionIndex` may be removed from the history if needed.
+    /// @param newRetentionIndex - new retention index
+    function pruneBottomUpCheckpoints(uint64 newRetentionIndex) external systemActorOnly {
```
aakoshh, line 316:
```suggestion
    /// @notice Set a new checkpoint retention height and garbage collect all checkpoints in range [`retentionHeight`, `newRetentionHeight`)
    /// @dev `retentionHeight` is the height of the first incomplete checkpointswe must keep to implement checkpointing.
    /// All checkpoints with a height less than `retentionHeight` are removed from the history, assuming they are committed to the parent.
    /// @param newRetentionHeight - the height of the oldest checkpoint to keep
    function pruneBottomUpCheckpoints(uint64 newRetentionHeight) external systemActorOnly {
```

When I see "index" I think of an array, but here it's the block height, isn't it? Using height is one less concept to remember.

src/gateway/GatewayGetterFacet.sol
aakoshh, line 226:
```suggestion
    /// @notice Calculate the threshold required for quorum in this subnet based on the configured majority percentage and the total weight of the validators.
    function getQuorumThreshold(uint256 totalWeight) public view returns (uint256) {
```

src/gateway/GatewayRouterFacet.sol
```diff
@@ -285,4 +308,28 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+        uint64 oldRetentionIndex = s.bottomUpCheckpointRetentionIndex;
+
+        if (newRetentionIndex <= oldRetentionIndex) {
+            revert InvalidRetentionIndex();
+        }
+
+        for (uint64 h = oldRetentionIndex; h < newRetentionIndex; ) {
+            delete s.bottomUpCheckpoints[h];
+            delete s.bottomUpCheckpointInfo[h];
+            delete s.bottomUpCollectedSignatures[h];
```
aakoshh, line 326: We should also clear the `bottomUpCheckpointsLegacy` while it exists.

src/gateway/GatewayGetterFacet.sol
dnkolegov: Fixed. That was to calculate the weight needed to complete a checkpoint.

src/gateway/GatewayRouterFacet.sol
dnkolegov: `bottomUpCollectedSignatures` is `mapping(uint64 => EnumerableSet.AddressSet)`
When we add something to an enumerable set, we get a boolean value, meaning success. If you ignore it, then linters will complain. This revert is equal to panicking because we must be able to add the signer.

I agree that the error is misleading. Fixed.

src/gateway/GatewayRouterFacet.sol
dnkolegov: The explanation here is the same as above.

src/gateway/GatewayGetterFacet.sol
dnkolegov: Applied. Thank you. Looks perfect.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Those two bottom-up checkpoint types are probably misleading. But `bottomUpCheckpointsLegacy` is not used in implementing the new bottom-up checkpoint mechanism. It is in the code just to build it without removing all libraries, etc.

src/gateway/GatewayGetterFacet.sol
```diff
@@ -194,4 +196,35 @@ contract GatewayGetterFacet {
...
+            checkpoints[i] = s.bottomUpCheckpoints[uint64(heights[i])];
+            unchecked {
+                ++i;
+            }
+        }
+        return checkpoints;
+    }
+
+    /// @notice get the bottom-up checkpoint retention index
+    function getBottomUpRetentionIndex() public view returns (uint64) {
```
aakoshh, line 221:
```suggestion
    function getBottomUpRetentionHeight() public view returns (uint64) {
```

src/gateway/GatewayRouterFacet.sol
dnkolegov: It will be fixed in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/214

aakoshh (approved):
Cool, cheers!

PR #208, opened 2023-09-22
Staking

cryptoAtwill:
Initial staking implementation as outlined in: https://github.com/consensus-shipyard/ipc-solidity-actors/issues/203#issuecomment-1729493531


src/structs/Subnet.sol
```diff
@@ -24,6 +24,78 @@ struct Subnet {
...
+/// Active validators are those that are producing blocks in the child subnet.
+/// Waiting validators are those that do no have as high collateral as Active validators.
+///
+/// The max number of active validators is limited by `activeLimit` and the size of waiting
+/// validators is not bounded.
+///
+/// With each validator staking change, waiting validators can be promoted to active validators
+/// and active validators can be knocked off.
+///
+/// @NOTE THAT THE LIST OF ACTIVE VALIDATORS ARE MAINTAINED OFFCHAIN DUE TO GAS CONCERNS!
```
aakoshh, line 89: Where is this supposed to be maintained? Could we not stick to the simplest on-chain solution and worry about concerns after we have developed a correct implementation?

cryptoAtwill: Yeah, that's possible. Just expensive.

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    /// @notice Confirm the changes in bottom up checkpoint submission, only call this in bottom up checkpoint execution.
+    function confirmChange() internal {
+        SubnetActorStorage storage s = LibSubnetActorStorage.appStorage();
+        StakingChangeSet storage changeSet = s.changeSet;
+
+        uint64 start = changeSet.startConfigurationNumber();
+        uint64 end = changeSet.nextConfigurationNumber;
```
aakoshh, line 135: You will not necessarily be able to confirm right to the top, only up to the point indicated by the bottom-up checkpoint.

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,153 @@
...
+        for (uint64 i = start; i < end; ) {
+            StakingChange storage change = changeSet.getChange(i);
+            address validator = change.validator;
+            uint256 amount = change.amount;
+
+            if (change.op == StakingOperation.Withdraw) {
+                s.validatorSet.withdraw(validator, amount);
+                s.releaseQueue.addNewRelease(validator, amount);
```
aakoshh, line 143: Where does it check that the validator indeed had this collateral and did not try to double withdraw?

aakoshh: Ok I see there is something [here](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/208/files#diff-8e6cc37f8b7f72be690d950e93e34c1acf79210f70785dc0504db8ded8399d20R122-R126)

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,153 @@
...
+
+        s.changeSet.upsertDeposit(validator, amount);
+        IGateway(s.ipcGatewayAddr).addStake{value: amount}();
+    }
+
+    /// @notice Withdraw `msg.value` amount of collateral by `msg.sender`
+    function withdraw() internal {
+        SubnetActorStorage storage s = LibSubnetActorStorage.appStorage();
+
+        uint256 totalWithdraw = s.changeSet.upsertWithdraw(msg.sender, msg.value);
```
aakoshh, line 122: I don't think this is safe. You are updating a configuration number that is potentially already being applied on the child subnet. When it reports it has confirmed something up to configuration number 5, the parent and the child could have a different idea of what that means.

aakoshh (commented):
Thank you again for taking the initiative on this. 

I can only repeat that I find the changset-only design difficult to reason about, compared to the one I proposed with a pending balance table and the active collateral updated after confirmations. 

At first glance I would say the "upsert" is an absolute no-go, the changesets are essentially top-down messages and should be immutable. If you think they are some kind of a staging area for malleable changesets vs changesets that have been sent to the child subnet, it's not clear in this PR and we should have a design document. 

I'll also repeat that having a design doc to capture the invariants of the system and the messages going between parent and child subnet would be a better start than the Solidity interface, because it could allow developing [state machine tests](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/testing/src/smt.rs) independently of the implementation, which would be the only objective measure.

src/structs/Subnet.sol
aakoshh: Maybe but the computation will happen somewhere regardless, we would just be hiding it. I'm just saying this is already very confusing and haphazard, mixing in off-chain and Fendermint will just make it more difficult to reason about.

src/lib/LibStaking.sol
cryptoAtwill: I have updated the design to include this change.

src/lib/LibStaking.sol
cryptoAtwill: I removed them

src/structs/Subnet.sol
cryptoAtwill: I think we can rule out sorting algorithms on chain as O(nlogn) can potentially be gas costly. Initially, I thought a `priority queue` would be enough. But updating non-min validator collateral and reheapify is also expensive. Then we might have to use something like a tree. It's complicated and expensive as well.

It is generally not recommended to maintain sorted data structure on chain. The recommended is usually store it unordered on chain and maintain order offchain.

Why not we store the active validators in the Gateway in the child net. So far I don't see any need to maintain active validators in the parent, we just need their collaterals. We have to calculate the validator set changes anyways in `fendermint`, we can query the last validator set with configuration number in the child, then with new change, we apply the changes in `fendermint`, if there are active validator change, we commit to child/fendermint.

Actually I dont see we need to query active validators in solidity in child. I could be wrong, but seems we dont have this use case.

aakoshh: The use case where we needed the active validators on the parent is what I commented about earlier: validating bottom-up checkpoint signatures.

aakoshh: What I am not getting is what do you mean by “off-chain” in our setup. 

In a normal dApp I understand, you have an application in the cloud and some part of it is posted on chain. You have to pay ridiculous amounts for gas, so you want to minimise _computation done by the validators_ by doing it yourself, storing the data in your local database, and sending commitments. 

In our case, however, Fendermint is all run by the validators, regardless of it’s done in Solidity or Rust, except the actors have persistent storage and the Rust part generally doesn’t. 

So, if you say gas is expensive because of the inherent cost of maintaining a priority queue, then it’s going to be expensive even if it’s pushed out on the host, except now it’s a hidden cost.

aakoshh: If we send the collaterals as-is to the child then one extra tool I see is that we could use implicit execution to make applying configurations free. Then we have to post back the changes so the parent can release collateral as well.

cryptoAtwill: Sorry, should have been more specific, `off-chain` means `off-parent`, that means maintaining the list in child.

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,233 @@
...
+
+    /// @notice Perform upsert operation to the withdraw changes, return total value to withdraw
+    /// @notice of the validator.
+    /// Each insert will increment the configuration number by 1, update will not.
+    function withdrawRequest(StakingChangeSet storage changes, address validator, uint256 amount) internal {
+        uint64 configurationNumber = changes.nextConfigurationNumber;
+        changes.nextConfigurationNumber = configurationNumber + 1;
+
+        emit NewStakingRequest({
+            op: StakingOperation.Deposit,
```
adlrocha, line 21: Is this a typo?
```suggestion
            op: StakingOperation.Withdraw,
```

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,233 @@
...
+        });
+    }
+
+    /// @notice Perform upsert operation to the deposit changes
+    function depositRequest(StakingChangeSet storage changes, address validator, uint256 amount) internal {
+        uint64 configurationNumber = changes.nextConfigurationNumber;
+        changes.nextConfigurationNumber = configurationNumber + 1;
+
+        emit NewStakingRequest({
+            op: StakingOperation.Withdraw,
```
adlrocha, line 34: Now I think I may be misunderstanding something, but is this also a typo?
```suggestion
            op: StakingOperation.Deposit,
```

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,233 @@
...
+
+    /// @notice Get the change at configuration number
+    function getChange(
+        StakingChangeSet storage changes,
+        uint64 configurationNumber
+    ) internal view returns (StakingChange storage) {
+        return changes.changes[configurationNumber];
+    }
+
+    function purgeChange(StakingChangeSet storage changes, uint64 configurationNumber) internal {
```
adlrocha, line 49: I get you need to pass the `changes` as an argument because libraries can have a state (although we may be able to avoid this by having a stateful library like here: https://dev.to/mudgen/solidity-libraries-can-t-have-state-variables-oh-yes-they-can-3ke9 (??)). I am not saying we should do it now, but I was wondering if it may be worth.

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,233 @@
...
+    /// @notice in ascending order. This method does not do checks on this.
+    function push(AddressStakingReleases storage self, StakingRelease memory release) internal {
+        uint16 length = self.length;
+        uint16 nextIdx = self.startIdx + length;
+
+        self.releases[nextIdx] = release;
+        self.length = length + 1;
+    }
+
+    /// @notice Perform compaction on releases, i.e. aggregates the amount that can be release
```
adlrocha, line 65:
```suggestion
    /// @notice Perform compaction on releases, i.e. aggregates the amount that can be released
```

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,233 @@
...
+        StakingRelease memory release = StakingRelease({releaseAt: releaseAt, amount: amount});
+
+        self.releases[validator].push(release);
+
+        emit NewCollateralRelease({validator: validator, amount: amount, releaseBlock: releaseAt});
+    }
+
+    /// @notice Validator claim the available collateral that are released
+    function claim(StakingReleaseQueue storage self, address validator) internal {
+        (uint256 amount, uint16 newLength) = self.releases[validator].compact();
```
adlrocha, line 124: This assumes that whomever is claiming may pay the gas of compacting the whole data structure for all the other queued validators, right? I don't how comfortable do we feel with this.

src/lib/LibStaking.sol
```diff
@@ -0,0 +1,233 @@
...
+        for (uint64 i = start; i <= configurationNumber; ) {
+            StakingChange storage change = changeSet.getChange(i);
+            address validator = change.validator;
+            uint256 amount = change.amount;
+
+            if (change.op == StakingOperation.Withdraw) {
+                s.validatorSet.confirmChange({validator: validator, amount: amount, op: StakingOperation.Withdraw});
+                s.releaseQueue.addNewRelease(validator, amount);
+            } else {
+                s.validatorSet.confirmChange({validator: validator, amount: amount, op: StakingOperation.Deposit});
```
adlrocha, line 218: Why are deposit changes queued? I would expect those to be able to go through immediately without having to wait for a bottom up checkpoint, right?

adlrocha (commented):
While all of the implementation makes sense, I would like to see the library integrated to the external functions to see if makes sense. I feel there are a lot of redundant things like the `changeSet` and `validatorSet` that could be reconciled into a single structure, and a few things (but maybe I am wrong). 

Would you mind taking a stab on integrating this to `join/leave` in a follow-up PR so the end-to-end is a bit clearer? Thanks!

src/lib/LibStaking.sol
cryptoAtwill: typo

src/lib/LibStaking.sol
cryptoAtwill: I'm assuming only the validator will do the compaction, i.e. msg.sender == validator.

adlrocha (approved):
I think there are no outstanding comments. Thanks!

src/lib/LibStaking.sol
cryptoAtwill: Yeah, that is possible. My past experience is as long as we are not paying the gas, it's ok, haha.

PR #209, opened 2023-09-26
Adding min priority queue

cryptoAtwill:
Adding min heap as priority queue implementation. Following this implementation: https://algs4.cs.princeton.edu/24pq/IndexMinPQ.java.html


src/lib/priority/LibMinPQ.sol
```diff
@@ -0,0 +1,163 @@
...
+        sink(self, validators, 1, value);
+    }
+
+    /// @notice Reheapify the heap when the validator is deleted.
+    /// NOTE that caller should ensure the queue is not empty.
+    function deleteReheapify(MinPQ storage self, ValidatorSet storage validators, address validator) internal {
+        uint16 pos = self.inner.addressToPos[validator];
+        uint16 size = self.inner.size;
+
+        self.inner.exchange(size, size);
```
aakoshh, line 59:
```suggestion
        self.inner.exchange(pos, size);
```

src/lib/priority/LibPQ.sol
```diff
@@ -0,0 +1,59 @@
...
+        return self.addressToPos[validator] != 0;
+    }
+
+    function del(PQ storage self, uint16 pos) internal {
+        address addr = self.posToAddress[pos];
+        delete self.posToAddress[pos];
+        delete self.addressToPos[addr];
+    }
+
+    function getCollateral(
```
aakoshh, line 40: It is confusing that the name of this method not `getConfirmedCollateral`; at first glance it sounds like there are two kinds of collaterals, but then it turns out it's the same.

src/lib/priority/LibMinPQ.sol
```diff
@@ -0,0 +1,163 @@
...
+        uint16 pos = self.inner.addressToPos[validator];
+        uint256 val = validators.getConfirmedCollateral(validator);
+        swim(self, validators, pos, val);
+    }
+
+    /// @notice Get the minimal value in the priority queue.
+    /// NOTE that caller should ensure the queue is not empty!
+    function min(MinPQ storage self, ValidatorSet storage validators) internal view returns (address, uint256) {
+        address addr = self.inner.posToAddress[1];
+        uint256 collateral = validators.getConfirmedCollateral(addr);
```
aakoshh, line 94: I don't know what to do about it but it's a bit leaky that in theory the collateral could change in the `validators` while the position is fixed in the priority queue. How do we know min is still min?

aakoshh: I am guessing that you will have something that owns both and makes sure they only get changed at the same time?

aakoshh (commented):
I didn't try to understand the implementation, but this is great stuff!

src/lib/priority/LibPQ.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.8.19;
+
+import {LibValidatorSet} from "../LibStaking.sol";
+import {ValidatorSet} from "../../structs/Subnet.sol";
+
+/// The implementation that mimics that in https://algs4.cs.princeton.edu/24pq/IndexMinPQ.java.html.
```
adlrocha, line 7:
```suggestion
/// The implementation that mimics the Java impl in https://algs4.cs.princeton.edu/24pq/IndexMinPQ.java.html.
```

src/lib/priority/LibPQ.sol
adlrocha: +1

src/lib/priority/LibMinPQ.sol
```diff
@@ -0,0 +1,163 @@
...
+        self.inner.posToAddress[size] = validator;
+
+        self.inner.size = size;
+
+        uint256 confirmedCollateral = validators.getConfirmedCollateral(validator);
+        swim(self, validators, size, confirmedCollateral);
+    }
+
+    /// @notice Pop the minimal value in the priority queue.
+    /// NOTE that caller should ensure the queue is not empty!
```
adlrocha, line 40: Should we add an `isEmpty` function like in the original implementation as a sanity-check? I am just worried we may face unexpected bugs :)

adlrocha (approved):
LGTM. I haven't looked in depth into the implementation, assuming that it is a translation of the Java impl :) If the tests pass I am happy merging this one. Thanks!

src/lib/priority/LibMinPQ.sol
cryptoAtwill: yep, it's actually the `ValidatorSet` that owns both of them.

src/lib/priority/LibMinPQ.sol
cryptoAtwill: I'm thinking adding these changes to the caller, so that we can save a few checks here. But I think adding them should be fine!

PR #211, opened 2023-09-27
Update rust binding

cryptoAtwill:
Rust binding in dev is outdated. Adding auto commit in github action.


adlrocha (approved):
LGTM. Thanks!

Just one quick concern, are we sure that the Rust compilation is deterministic? I remember we were having some issues with bindings changing between compilation even if the code hadn't changed. I am asking because this could lead to a lot of noise with bindings changing in every PR.

cryptoAtwill:
@adlrocha from my past experience, without reading the compilation source code or ethers, I cannot say the binding is deterministic. But this way at least we won't have to worry about binding up to date.

.github/workflows/test.yml
```diff
@@ -12,6 +12,7 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
aakoshh, line 15: Why is this needed?

cryptoAtwill: Otherwise cannot push as the HEAD is not attached to any branch

aakoshh (approved):
Nice one :+1:

.github/workflows/test.yml
aakoshh: Thanks, it's worth adding a comment about this I reckon

PR #212, opened 2023-09-27
Integrate pq

cryptoAtwill:
Integrate priority queue into staking. 

We use `min-heap` to maintain the active validators, and `max-heap` to maintain the waiting validators. When there is a change, we update the membership using pop and insert of each queue.


aakoshh (approved):
This looks great! 

I didn't read the tests, but I see there is a lot of them :man_dancing: 

I'd like to understand the flow a little bit more, though. So we say that the active/waiting validator set changes when deposits and withdrawals are _confirmed_. What does _confirm_ mean exactly?

Does it mean that the bottom-up checkpoint indicated this change has been part of a configuration number?

If so, do you expect that the child subnet will do a parallel calculation of the exact same nature to calculate who the active validators are? That is, the child subnet will not receive changesets that tell absolute power levels of the active validator set, they will get all deposits and withdrawals and are expected to do their own calculation which should match what is going on here, right?

This is where I thought we'd have two instances of the validator sets, the _pending_ and the _current_, so we can tell the child what values it should do. 

So, with this we'll replicate the entire validator collateral table on the child by sending deltas, which might be a result of a delegation (if we have delegations), and then it's up to us whether we want to a) retrieve the entire table plus the changes and do an ad-hoc  top N calculation in Rust at the end of the checkpoint period or b) replicate this machinery in Solidity.

cryptoAtwill:
@aakoshh By confirmed it means bottom up checkpoint submission confirms the configuration number change.

Regarding propagating the change set, my understanding is aligned with [this](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1695818870547679?thread_ts=1695814082.418389&cid=C04JR5R1UL8). When it comes to implementation, my understanding is:
- Parent syncer in fendermint child polls the change set.

(If we rely on solidity)
- 1: Pushes these changes to child gateway. Child gateway just stores the changes. 
- 2: Bottom up checkpoint submission pulls the changes from the gateway, submits the changes to child. Child gateway uses the same `ValidatorSet` as SubnetActor, applies the changes to update the absolute powers.
- 3: The changes in absolute powers can be obtained and sent to cometbft.
- 4: Bottom up checkpoint confirms configuration number to parent subnet actor.

(If we rely on fendermint)
- 1: Pushes these changes to child gateway. Child gateway just stores the changes. 
- 2: Bottom up checkpoint submission pulls the changes from the gateway. Fendermint applies the changes to update the absolute powers.
- 3: The updates in absolute power are committed directly to child gateway and sent to cometbft.
- 4: Bottom up checkpoint confirms configuration number to parent subnet actor.

IMO, I think relying on `fendermint` would be (slightly) faster as it directly calculates the changes without going to fevm. Also in fendermint, it's like a offchain process which does not involve gas. But relying on `Solidity` will be simpler to maintain as we only maintain one code base for validator updates. Probably as our first try we can just use `solidity` to quicker iteration?

Also, I'm curious for step 1 -> step 2, can't we skip pushing to gateway and directly fetch the changes from parent syncer? I mean pushing to gateway is less error prone, just a thought on the possibilities.

aakoshh:
>  1: Pushes these changes to child gateway. Child gateway just stores the changes.

We'll need to see a PR that defines what exactly the definition of a "change" is. I am guessing that it means a +/- on _any_ validator, not just the top N?

> (If we rely on fendermint)
> 2: Bottom up checkpoint submission pulls the changes from the gateway. Fendermint applies the changes to update the absolute powers.

Fendermint will have to pull the changes as well as the current balances of all waiting validators to calculate who might be entering the active validator set. I can only ask CometBFT for the active set.

> 3: The updates in absolute power are committed directly to child gateway and sent to cometbft.

Agreed, any change in the balances have to be committed back.

> I think relying on fendermint would be (slightly) faster as it directly calculates the changes without going to fevm.

I have no idea to be honest, because all that data has to be moved to Fendermint and a minimums/maximums found.

>  Probably as our first try we can just use solidity to quicker iteration?

Why not? In that case there would need to be a method to call at the end of the period which applies all the changes and could return either:
* The new active validator set, similarly to how it returns the full membership today. Fendermint can calculate the delta with the one already in CometBFT to say anyone in the old but not in the new went to 0 power.
* Just the the validators whose power changed, including the ones who went to 0 due to leaving the active validator set.

> can't we skip pushing to gateway and directly fetch the changes from parent syncer?

Good question! I had to think for a minute; it helps that I just did https://github.com/consensus-shipyard/fendermint/pull/289 yesterday. 

Okay so you are asking why we are depositing the changes in the gateway and why we don't go fetch it live from the parent node when we apply the changes in the child. 

Semantically I don't think there would be a difference in the result, but it could be terrible in terms of variance of block processing time. 

Currently we do this:
1. Validators periodically determine what is the latest finalized height on the parent. 
2. Whoever is proposing a block can include a proposal to move forward with the parent finalized height. 
3. If the others vote for it, then they all fetch the limited number of blocks since the last finalized height, which includes getting i) the cross messages as well as ii) the validator updates. 
4. Validator updates are stashed in the gateway
5. Cross messages are executed (or stashed and executed in each subsequent block, unless Lotus).
6. If the block height is the epoch boundary, we retrieve the validator updates from the gateway and update CometBFT.

One thing to clarify is that the question is not about executing validator updates as soon as they appear, because then we wouldn't be adhering to the bottom-up checkpoint period, that is, we should not change who creates blocks and signs checkpoints until the period ends. Instead we are just asking if 3.ii) can be skipped and deferred to step 6)

The problem comes up if we think about some extreme values we could use. What if we want a really long checkpointing period? If every node started querying thousands of blocks on the parent to look for validator set changes in the emitted events, right in the middle of block processing, it could easily cause all sorts of timeouts in CometBFT and affect liveness. Ostensibly we could make a query that returns all changes in one method call from the parent, but how large can this result be? 

True, if someone did not vote for the finality of a parent block but it still reached quorum, they will have no choice but to fetch it during block processing, but we should strive to make this the exception, rather than the rule. By always deferring, validators would have no choice. 

So, by doing the fetching at times of finality, we are chipping away at the problem and keeping the IO naturally limited.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -200,4 +201,62 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+    function setData(bytes calldata data) external {
+        if (!LibStaking.hasStaked(msg.sender)) {
+            revert NotStakedBefore();
+        }
+        LibStaking.setValidatorData(msg.sender, data);
+    }
+
+    /// @notice method that allows a validator to join the subnet
+    /// @param data The offchain data that should be associated with the validator
+    function join2(bytes calldata data) external payable notKilled {
```
adlrocha, line 215: This `join2` will replace the current `join` function, right? They won't coexist.

src/structs/Subnet.sol
```diff
@@ -97,10 +99,14 @@ struct Validator {
...
     /// The total number of active validators allowed.
     uint16 activeLimit;
-    /// The total number of validators currently in the set.
-    uint16 totalValidators;
+    /// The total collateral confirmed.
+    uint256 totalConfirmedCollateral;
     /// The mapping of each validator address to its information.
     mapping(address => Validator) validators;
+    /// @notice The active validators tracked using min priority queue.
+    MinPQ activeValidators;
```
adlrocha, line 107: Reading this code I realized it may help to have a note (either here or in their implementation) of the data structure stored in the priority queue, and the value used to determine their order. It is not obvious from the name, and we are not using a type-agnostic implementation for the priority queues, so we should be a bit more explicit to ease the understanding of the code.

src/lib/LibStaking.sol
```diff
@@ -142,6 +158,19 @@ library LibValidatorSet {
         collateral = validators.validators[validator].confirmedCollateral;
     }
 
+    function isActiveValidator(ValidatorSet storage self, address validator) internal view returns (bool) {
+        return self.activeValidators.contains(validator);
+    }
+
+    /// @notice Set validator data
+    function setData(ValidatorSet storage validators, address validator, bytes calldata data) internal {
```
adlrocha, line 166: It would useful to give an example of what we expect to store in `data`. Is this the worker address, the network address, any arbitrary metadata?

src/lib/LibStaking.sol
```diff
@@ -158,21 +187,171 @@ library LibValidatorSet {
         validators.validators[validator].totalCollateral = total;
     }
 
-    function confirmDeposit(ValidatorSet storage validators, address validator, uint256 amount) internal {
-        validators.validators[validator].confirmedCollateral += amount;
+    function confirmDeposit(ValidatorSet storage self, address validator, uint256 amount) internal {
```
adlrocha, line 190: When is this function used?

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -200,4 +201,62 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        if (msg.value == 0) {
+            revert CollateralIsZero();
+        }
+
+        LibStaking.setValidatorData(msg.sender, data);
+        LibStaking.deposit(msg.sender, msg.value);
+    }
+
+    /// @notice method that allows a validator to increase their stake
+    function stake() external payable notKilled {
```
adlrocha, line 225: This is something I've been looking to do for a while, instead of having a single `join` function that does both, registering a validator and adding stake, to have two independent methods for this. Thanks!

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -200,4 +201,62 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+    /// @notice Set the data of a validator
+    function setData(bytes calldata data) external {
+        if (!LibStaking.hasStaked(msg.sender)) {
+            revert NotStakedBefore();
+        }
+        LibStaking.setValidatorData(msg.sender, data);
+    }
+
+    /// @notice method that allows a validator to join the subnet
+    /// @param data The offchain data that should be associated with the validator
```
adlrocha, line 214: We should be more explicit about the offchain data expected here. I like the idea of this being subnet-specific, but for the reference implementation we would need at least the `net-address` and the `worker-address`, right? 

This is what @aakoshh needs to track the validator's public keys, and we will have to update the `ipc-cli` accordingly.

adlrocha (approved):
Amazing job with this one, @cryptoAtwill. I left a few comments (maybe due to my lack of context) but overall LGTM. 

The only thing left would be to remove the existing `join` and `leave` functions to replace them by `join2` and `leave2`, right? Also, let's double-check that we haven't left any relevant check in any of those functions that should be included in the new ones (just in case).

src/lib/LibStaking.sol
adlrocha: Also, I think we should call it `metadata` instead of `data` if possible.

src/subnet/SubnetActorManagerFacet.sol
cryptoAtwill: yep, will replace. I just afraid replacing would create too many conflicts, so I created `join2`.

PR #214, opened 2023-09-28
Use new bottom-up checkpoint type

dnkolegov:
This PR addresses the needs of https://github.com/consensus-shipyard/fendermint/pull/286.

- To get a bottom-up checkpoint, use `Getter.bottomUpCheckpoint(height)`
- to get cross-messages, use  `Getter.bottomUpMessages(height)`


src/structs/Checkpoint.sol
```diff
@@ -41,6 +41,8 @@ struct BottomUpCheckpointNew {
     /// approach where the messages are fetched with Bitswap and provided by Fendermint, or the full-fat
     /// approach we need with Lotus, where the messages are part of the relayed transaction.
     bytes32 crossMessagesHash;
+    /// @def Fee is a legacy field. Should be removed when we have the fee payment figured out in the execution use the new model.
+    uint256 fee;
```
aakoshh, line 45: I don't think we need to add this. We already know we're not going to use it for paying for gas, and it's not clear what it should be used for. I can sum up the fees but if we don't use it then what's the point?

src/lib/LibGatewayActorStorage.sol
```diff
@@ -31,13 +31,15 @@ struct GatewayActorStorage {
...
     /// @notice A mapping of block numbers to bottom-up checkpoints
     // slither-disable-next-line uninitialized-state
-    mapping(uint64 => BottomUpCheckpointNew) bottomUpCheckpoints;
+    mapping(uint64 => BottomUpCheckpoint) bottomUpCheckpoints;
     /// @notice A mapping of block numbers to checkpoint data
     // slither-disable-next-line uninitialized-state
     mapping(uint64 => CheckpointInfo) bottomUpCheckpointInfo;
+    /// @notice A mapping of block numbers to bottom-up cross-messages
+    // slither-disable-next-line uninitialized-state
+    mapping(uint64 => CrossMsg[]) bottomUpMessages;
```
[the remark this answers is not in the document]
dnkolegov: Fixed

src/structs/Checkpoint.sol
dnkolegov: Fixed

aakoshh (approved):
Smashing!

src/lib/LibGateway.sol
```diff
@@ -39,8 +38,8 @@ library LibGateway {
     {
         GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
         epoch = LibVoting.getNextEpoch(block.number, s.bottomUpCheckPeriod);
-        checkpoint = s.bottomUpCheckpointsLegacy[epoch];
-        exists = !checkpoint.source.isEmpty();
+        checkpoint = s.bottomUpCheckpoints[epoch];
```
aakoshh, line 41: NB this is a little awkward: the checkpoint is created here, but because its fields are left empty it's like a non-existing value, right? I just checked to make sure https://github.com/consensus-shipyard/ipc-solidity-actors/blob/addbc7e92691cf30894af2cdb0bf4a72de32ce13/src/gateway/GatewayRouterFacet.sol#L221 won't reject it when it's eventually being added, and I think it's okay, but semantically the `getCurrentBottomUpCheckpoint` no longer makes sense.

dnkolegov: Why do you think it is created here?

> getCurrentBottomUpCheckpoint no longer makes sense.
It is not obvious because it is used in other places of the actors. E.g. in `release` external method.

dnkolegov: Not relevant. We are removing libvoting

aakoshh: > Why do you think it is created here?

I am not a Solidity expert but I thought it was created because of how it worked before:

In the `getCurrentBottomUpCheckpoint`:
```solidity
       checkpoint = s.bottomUpCheckpointsLegacy[epoch];
       exists = !checkpoint.subnetID.isEmpty();
```
then in `commitBottomUpMsg`:
```solidity
        (, , BottomUpCheckpoint storage checkpoint) = getCurrentBottomUpCheckpoint();
        checkpoint.crossMsgs.push(crossMessage);
```
So it ignored `exists` and appended the message anyway. Unless it was impossible for this thing not to exist when `commitBottomUpMsg` was called, I just assumed by pushing things into it it will have been created. At least that's how I imagined it worked and I asked you about it in other PRs: we get a record from the collection and seemingly are free to modify its fields without having to put them back. 

This could be a total misunderstanding on my part.

aakoshh: For example below it says `s.bottomUpMessages[epoch].push(crossMessage);` and if getting the record from the collection did not "upsert" it then the pushed message would be lost.

PR #215, opened 2023-09-29
Add a coverage test

dnkolegov:
This PR adds a simple tool to measure test coverage by running `make coverage` and running the same as a workflow on CI/CD.

The current threshold is **10** %. If the coverage of line, blocks, or functions is low, then the threshold then the workflow panics.

Screenshot 2023-09-29 at 11 38 16


adlrocha (approved):
This is really useful. Thanks!

Just a few questions: 
- The pipeline checks the coverage threshold for all the lines of code in the project, right? Not for the ones changed in the PRs. This means that before we enforce this check we should have a baseline threshold over the one we currently have because if not CI will always be unhappy. The issue with this is that if we set the threshold initially too low, new PRs may only cover that small threshold. In any case, this is useful, and we can merge and try to dedicate a sprint to improve coverage.
- Also, I guess the new bindings included in the repo have been auto-generated, right? I am asking to understand if this may be an issue: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/211#pullrequestreview-1652469864

dnkolegov:
> The pipeline checks the coverage threshold for all the lines of code in the project, right? 

Correct. I believe it is very easy to find that threshold, especially in our case. The main idea is to be sure we do not decrease the coverage by adding new code.

> Also, I guess the new bindings included in the repo have been auto-generated, right?

Will fix.

dnkolegov:
https://github.com/consensus-shipyard/ipc-solidity-actors/issues/217

PR #216, opened 2023-09-29
Remove libvoting

dnkolegov:
This PR removes the unused LibVoting library and the relevant types.


.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,31 @@
+name: Coverage check
```
adlrocha, line 1: Why is this included in this PR instead of https://github.com/consensus-shipyard/ipc-solidity-actors/pull/215?

src/gateway/GatewayManagerFacet.sol
```diff
@@ -14,21 +14,18 @@ import {SubnetIDHelper} from "../lib/SubnetIDHelper.sol";
...
 
 contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
     using FilAddress for address payable;
     using SubnetIDHelper for SubnetID;
     using FvmAddressHelper for FvmAddress;
 
     /// @notice initialize the contract with the genesis epoch
-    /// @param genesisEpoch - genesis epoch to set
-    function initGenesisEpoch(uint64 genesisEpoch) external systemActorOnly {
+    function initGenesisEpoch() external systemActorOnly {
```
adlrocha, line 24: @cryptoAtwill, do we still need this function for anything? I would say we can remove it (I actually removed it myself on this PR that has been superseded: #199). 

Btw, @dnkolegov, do you mind removing all the `genesis` field as I did in that PR as part of this one? That way we have everything in the same PR. Thanks!

src/interfaces/IGateway.sol
```diff
@@ -66,4 +62,22 @@ interface IGateway {
         FvmAddress[] calldata validators,
         uint256[] calldata weights
     ) external;
+
+    /// @notice creates a new bottom-up checkpoint
+    function createBottomUpCheckpoint(
```
adlrocha, line 67: These are for the new checkpoint implementation, right? Wasn't this included in some other PR?

adlrocha (changes requested):
I love how we are removing a lot of outdated code with this one, thanks!

I left a few requests to see if we can remove a few more things from this PRs, but overall looks good. 

One outstanding thing that I haven't seen is an update of the deployment scripts removing the deployment of `LibVoting`, but I don't see it in the script so I guess that we already did it after the huge refactor. Thanks!

src/interfaces/IGateway.sol
dnkolegov: This is for the new checkpoint. It was not.

src/gateway/GatewayManagerFacet.sol
dnkolegov: Removed https://github.com/consensus-shipyard/ipc-solidity-actors/pull/216/commits/f7a7a4253455ca7e7a5656931862d2c3b0d44129

src/interfaces/IGateway.sol
adlrocha: Gotcha, let's try to make PRs self-contained (i.e. I would have put this one in some other PRs to make reviews easier to folow). Thanks!

src/gateway/GatewayManagerFacet.sol
```diff
@@ -49,7 +37,6 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         subnet.id = subnetId;
         subnet.stake = msg.value;
         subnet.status = Status.Active;
-        subnet.genesisEpoch = block.number;
```
adlrocha, line 52: Do not remove this one: `genesisEpoch` need to be set when a subnet is activated.

src/interfaces/IGateway.sol
```diff
@@ -25,10 +25,6 @@ interface IGateway {
     /// hierarchy.
     function kill() external;
 
-    /// CommitChildCheck propagates the commitment of a checkpoint from a child subnet,
-    /// process the cross-messages directed to the subnet.
-    function commitChildCheck(BottomUpCheckpointLegacy calldata bottomupCheckpoint) external;
```
adlrocha, line 30: I think we should recover this function, it implements the common logic to trigger side-effects and execute cross-net messages in the parent. It may change slightly with the new approach but it needs to exist.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -170,34 +157,46 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        _commitCheckpoint({
+            checkpoint: checkpoint,
+            membershipRootHash: membershipRootHash,
+            membershipWeight: membershipWeight
+        });
     }
 
     /// @notice method that commits a checkpoint after reaching majority
     /// @param checkpoint - the batch messages data
-    function _commitCheckpoint(BottomUpCheckpointLegacy calldata checkpoint) internal {
```
adlrocha, line 192: Do not remove this one, this should still exist, it will be called when a relayer submits a valid checkpoint.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -170,34 +157,46 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
 
     /// @notice methods that allows a validator to submit a checkpoint (batch of messages) and vote for it with it's own voting power.
     /// @param checkpoint - the batch messages data
-    function submitCheckpoint(BottomUpCheckpointLegacy calldata checkpoint) external {
```
adlrocha, line 173: I don't know if there is already a function for submitting the checkpoint by the relayed with the array of signatures. If this doesn't exist we should keep this function.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -170,34 +157,46 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
 
     /// @notice method that commits a checkpoint after reaching majority
     /// @param checkpoint - the batch messages data
-    function _commitCheckpoint(BottomUpCheckpointLegacy calldata checkpoint) internal {
-        /// Ensures the checkpoints are chained. If not, should abort the current checkpoint.
-        if (s.prevExecutedCheckpointHash != checkpoint.prevHash) {
-            revert InconsistentPrevCheckpoint();
-        }
-
-        s.committedCheckpoints[checkpoint.epoch] = checkpoint;
```
adlrocha, line 198: @aakoshh, @dnkolegov, please, help me with this one. Do we need to store checkpoints in the parent in the new design? In the previous one we store them for future reference by users (and potential security operations on top of them). I personally would love to keep that one for now, but no strong opinion. 

(if this is the case we should keep this logic, and not only the child checkpoint commitment).

adlrocha (changes requested):
A few more comments before we merge.

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Restored

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Restored

src/interfaces/IGateway.sol
dnkolegov: Restored but it is unuseful now since all its code is disabled

adlrocha: All except sending `commitChildCheck` to the gateway should be disabled, yes.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -39,6 +39,81 @@ contract GatewayRouterFacet is GatewayActorModifiers {
     event QuorumReached(uint64 height, bytes32 checkpoint, uint256 quorumWeight);
     event QuorumWeightUpdated(uint64 height, bytes32 checkpoint, uint256 newWeight);
 
+    /// @notice submit a checkpoint in the gateway. Called from a subnet once the checkpoint is voted for and reaches majority
+    function commitChildCheck(BottomUpCheckpoint calldata commit) external {
```
adlrocha, line 43: Let's work on this one together next to wrap-up the execution of bottom-up checkpoints. I think we are getting close :)

adlrocha (approved):
I think now everything should be in place to merge. Let's tackle the execution of bottom-up next. Thanks!

PR #218, opened 2023-10-02
Getter for signatures

dnkolegov:
This PR adds a getter for signatures


src/lib/LibGatewayActorStorage.sol
```diff
@@ -48,8 +48,10 @@ struct GatewayActorStorage {
     /// @notice A list of incomplete checkpoints.
     // slither-disable-next-line uninitialized-state
     EnumerableSet.UintSet incompleteCheckpoints;
-    /// @notice The validators have already sent signatures at height `h`
-    mapping(uint64 => EnumerableSet.AddressSet) bottomUpCollectedSignatures;
+    /// @notice The addresses of the validators that have already sent signatures at height `h`
+    mapping(uint64 => EnumerableSet.AddressSet) bottomUpSignatureSenders;
```
adlrocha, line 52: I guess this is a trick to make the iteration of signature more gas efficient?

adlrocha (approved):
LGTM. Thanks!

src/lib/LibGatewayActorStorage.sol
dnkolegov: It is a trick just to extract them easily

PR #219, opened 2023-10-03
submitCheckpoint method

dnkolegov:
This is an implementation of the `submitBottomUpCheckpoint` in the parent subnet.


src/interfaces/ISubnetActor.sol
```diff
@@ -21,10 +21,11 @@ interface ISubnetActor {
     /// Tracks the accumulated rewards for each validator.
     function reward(uint256 amount) external;
 
-    /// SubmitCheckpoint accepts signed checkpoint votes for validators.
+    /// Executes the checkpoint if it is valid.
```
adlrocha, line 24:
```suggestion
    /// Executes the checkpoint if it is valid.
    /// It triggers the commitment of the checkpoint, 
    /// the execution of related cross-net messages, 
    /// and any other side-effects that need to be triggered
    /// by the checkpoint
```

src/lib/LibSubnetActorStorage.sol
```diff
@@ -21,8 +21,7 @@ struct SubnetActorStorage {
     mapping(address => FvmAddress) validatorWorkerAddresses;
     /// @notice contains all committed bottom-up checkpoint at specific epoch
     mapping(uint64 => BottomUpCheckpoint) committedCheckpoints;
-    /// @notice genesis block
-    bytes genesis;
+    uint64 lastBottomUpCheckpointExecutedHeight;
```
adlrocha, line 24: Weak opinion, but we may not need the `Executed` clarification, right? Any accepted checkpoint is now executed (we don't have voting like before).
```suggestion
    uint64 lastBottomUpCheckpointHeight;
```

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -158,47 +158,50 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+    /// @param messages The list of executed cross-messages
     function submitCheckpoint(
+        address[] calldata signatories,
         BottomUpCheckpoint calldata checkpoint,
-        bytes32 membershipRootHash,
-        uint256 membershipWeight
+        bytes calldata signatures,
+        CrossMsg[] calldata messages
     ) external {
+        if (checkpoint.blockHeight <= s.lastBottomUpCheckpointExecutedHeight) {
```
adlrocha, line 172: Probably this is a question also for @aakoshh. Do we want to allow gaps in the commitment of checkpoints? I don't think this should be the case as we may mess up with the execution of cross-net messages. 

If this is the case, we not only should make this check, but also ensure that the height being submitted is the immediate one following the previous checkpoint executed.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -158,47 +158,50 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+            revert HeightAlreadyExecuted();
+        }
+        if (checkpoint.blockHeight % s.bottomUpCheckPeriod != 0) {
+            revert InvalidCheckpointEpoch();
+        }
         if (s.status != Status.Active) {
```
adlrocha, line 178: @cryptoAtwill and I are planning to remove the status from the subnet actor, and have the gateway as the gatekeeper for the commitment of checkpoints and the one that enforces the minimum collateral in order to make cross-net operations, so this check will be deemed unnecessary. I will have a PR for that soon.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -158,47 +158,50 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
             revert SubnetNotActive();
         }
-        if (!s.validators.contains(msg.sender)) {
+        if (!s.validatorSet.isActiveValidator(msg.sender)) {
```
adlrocha, line 181: We can remove this check, now the checkpoint doesn't need to be sent by a validator. Anyone in the system can relay the signed checkpoint :champagne:

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -158,47 +158,50 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+    /// @notice method that commits a checkpoint after its validation and reaching majority
     /// @param checkpoint - the batch messages data
-    function _commitCheckpoint(
-        BottomUpCheckpoint calldata checkpoint,
-        bytes32 membershipRootHash,
-        uint256 membershipWeight
-    ) internal {
+    function _commitBottomUpCheckpoint(BottomUpCheckpoint calldata checkpoint) internal {
         s.committedCheckpoints[checkpoint.blockHeight] = checkpoint;
+        // TODO: remove this line and prevExecutedCheckpointHash ?
```
adlrocha, line 201: I personally like the idea of having checkpoint hash-linked even if their relationship is currently well-defined in the state of the contract, as it would allow us to make off-chain verifications over the state in the future. That being said, I agree with @aakoshh that they are currently useless so if they are giving us more headaches than anything else feel free to remove.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -158,47 +158,50 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
         s.prevExecutedCheckpointHash = checkpoint.toHash();
 
-        IGateway(s.ipcGatewayAddr).createBottomUpCheckpoint({
-            checkpoint: checkpoint,
-            membershipRootHash: membershipRootHash,
-            membershipWeight: membershipWeight
-        });
+        IGateway(s.ipcGatewayAddr).commitBottomUpCheckpoint(checkpoint);
```
adlrocha, line 204: This function from the gateway was called `commitChildCheck` before. I think it may useful to have the new implementation of that  function here too to have the end-to-end of this function (or in an immediate follow-up PR)

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -274,7 +277,8 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
      * @param hash The hash of the checkpoint.
      * @param signatures The packed signatures of the checkpoint.
      */
-    function validateCheckpoint(address[] memory signatories, bytes32 hash, bytes memory signatures) external view {
+    // TODO: should it be public or internal?
```
adlrocha, line 280: Good point, having it public may allow sanity-check verification by external users. Is there any additional costs of having it public?

adlrocha (commented):
Looking good. I left a few comments here and there, but I think we are in a good path. I feel the implementation will be way cleaner than before.

src/interfaces/IGateway.sol
```diff
@@ -27,7 +27,7 @@ interface IGateway {
 
     /// CommitChildCheck propagates the commitment of a checkpoint from a child subnet,
```
aakoshh, line 28:
```suggestion
    /// commitBottomUpCheckpoint propagates the commitment of a checkpoint from a child subnet,
```

src/interfaces/IGateway.sol
```diff
@@ -27,7 +27,7 @@ interface IGateway {
...
     /// process the cross-messages directed to the subnet.
-    function commitChildCheck(BottomUpCheckpoint calldata bottomUpCheckpoint) external;
+    function commitBottomUpCheckpoint(BottomUpCheckpoint calldata bottomUpCheckpoint) external;
```
aakoshh, line 30: Who is supposed to call this method? Should it be `internal`?

src/interfaces/ISubnetActor.sol
```diff
@@ -21,10 +21,11 @@ interface ISubnetActor {
...
     function submitCheckpoint(
+        address[] calldata signatories,
```
aakoshh, line 26: Minor but I would :100: put this next to the `signatures` parameter, rather than have `checkpoint` sandwiched between the two.

src/interfaces/ISubnetActor.sol
aakoshh: ...such as relayer reward book keeping

src/lib/LibSubnetActorStorage.sol
aakoshh: This should be the case with Lotus, but not necessarily with Fendermint if we use CIDs. In that case we would pass the checkpoint to Solidity twice:
1. The relayers submit the checkpoint to the contract without any messages, only the CID as opaque data over which the contract needs to check the signatures (NOTE: we need to revisit this because a CID is not 32 bytes, but the multi-sig validator expects a 32 byte hash, so we need to decide how to handle this: should we double hash the `cross_messages_hash` field before signing?)
2. The contract checks the quorum, and records the relayer for (later) rewards, as well as the submitted but _unexecuted_ checkpoint. There can be other submissions where only the relayer needs to be recorded.
3. Fendermint resolves the CID and executes the messages implicitly, marking the checkpoint as _executed_. 
4. If Fendermint is restarted before it has resolved the CID, it queries the contract for any _unexecuted_ checkpoints so that it can resume resolving the CID.

src/subnet/SubnetActorManagerFacet.sol
aakoshh: I agree that when checkpoints are submitted we have to check that their height is the next expected height according to the bottom up checkpoint period. I mentioned it in https://github.com/consensus-shipyard/fendermint/issues/192 but didn't expand on the rules. 

The reason we have to check is because the current validator set activate in checkpoint `N` is only eligible to sign the next checkpoint `N+1`, at which point _it might change the configuration number_, so we won't know who exactly should sign the checkpoint `N+2` after that, until we see checkpoint `N+1`. If we accepted `N+k` and validated it with the current validator set, it might turn out that they actually unbonded in `N+1` and should not have signed anything any longer.

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Cool, please tag me because I am curious how this will work out, e.g. how do you allow all validators to unbond if they are no longer interested in participating in the subnet (which would be the end of the subnet).

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Yes, this was exactly the thing we wanted to fix: if we have 100 validators we don't want to see all cross messages sent 100 times, with a different signature each time. Now it would be even worse: 100 times submitted with 100 signatures in each message. 

Instead, we want a handful of relayers to work for rewards - not too many, but also not so few that if one of them crashes the submissions stop completely.

src/lib/LibSubnetActorStorage.sol
aakoshh: > NOTE: we need to revisit this because a CID is not 32 bytes, but the multi-sig validator expects a 32 byte hash,

Never mind, I keep forgetting that we are signing the hash of the checkpoint, not the hash of the messages :facepalm:

src/subnet/SubnetActorManagerFacet.sol
aakoshh: I don't know how delegation works, a commit submitting to another commit... but where is the next configuration number applied?

src/interfaces/ISubnetActor.sol
dnkolegov: Ok

src/interfaces/IGateway.sol
dnkolegov: SubnetActor calls (called) it from `_commitBottomUpCheckpoint`. Because of that, it can not be `internal` unless we change the interaction between contracts.

src/interfaces/IGateway.sol
dnkolegov: Fixed

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: No, as far as I remember, public means it can be called externally and internally.

src/interfaces/IGateway.sol
aakoshh: Thanks, I didn't know that was how `internal` worked. So how do we make sure this isn't called by a user directly?

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Added a comment

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Removed for now. It can be restored when all types and interactions are defined.

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Another potential question is what should happen if we received an equivocating checkpoint.

Maybe that's something for @hmoniz to consider. If we had an evidence module for slashing then this could forward to that, or the users would have to make a second call.

src/subnet/SubnetActorManagerFacet.sol
aakoshh: I wouldn't make this public for the following reason:
* It says `validateCheckpoint` but really it validates any hash; it just happens that we call it with the hash of a checkpoint. 
* It validates it with the _current validator set_, which fits into our flow, but if you consider the method name, ostensibly it could validate it with the validator set active at the checkpointed height. In other words if you validate a checkpoint, apply it, then validate it again, it might turn invalid if the validator set changed. 
* It also doesn't do a full validation, only signature checks. 

So, at this point, because of how the method is named and how it works, it would be a confusing one for users to call. 

Maybe rename it to `validateActiveQuorum` or something. Users who want to query this can always go back to a previous block height, but at least there would be a hint of where the weights would be coming from.

src/lib/LibSubnetActorStorage.sol
dnkolegov: renamed

src/subnet/SubnetActorManagerFacet.sol
adlrocha: > That said: in the future we should not revert here, but recognise that we had a second or third relayer for the same checkpoint and record this fact, so that rewards can be distributed later, say, when the next checkpoint arrives. Then, we just return because don't have to execute (or enqueue) the messages again.

Agree. But for now I would suggest keeping it simple, just check that is the next immediate checkpoint, and let's track this on an issue.

src/lib/LibSubnetActorStorage.sol
adlrocha: We will have to do have a two-step execution of checkpoints once we have the low-fat version. But let's focus on the full-fat version for now (disregarding rewards) and let's build on top of that later.

src/subnet/SubnetActorManagerFacet.sol
adlrocha: I think I shared it on Slack yesterday: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/223. That being said, I currently don't tackle this issue of a subnet "going out of business for lack of validators".

src/subnet/SubnetActorManagerFacet.sol
adlrocha: You mean for the membership? This happens in the subnet actor. This is just to trigger the execution of the cross-net messages, as the "gatekeeper" of IPC is the gateway, it is the one enforcing the firewall requirement.

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Created https://github.com/consensus-shipyard/ipc-solidity-actors/issues/226

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -29,32 +29,42 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+     */
     function submitCheckpoint(
         BottomUpCheckpoint calldata checkpoint,
-        bytes32 membershipRootHash,
-        uint256 membershipWeight
+        CrossMsg[] calldata messages,
+        address[] calldata signatories,
+        bytes[] calldata signatures
     ) external {
+        // validations
```
aakoshh, line 46: It might be worth leaving a TODO here about relayer rewards. We havent' created any issue for it yet.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -29,32 +29,42 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
-            checkpoint: checkpoint,
-            membershipRootHash: membershipRootHash,
-            membershipWeight: membershipWeight
-        });
+        // effects
+        s.committedCheckpoints[checkpoint.blockHeight] = checkpoint;
+        s.lastBottomUpCheckpointHeight = checkpoint.blockHeight;
 
-        LibStaking.confirmChange(checkpoint.nextConfigurationNumber);
+        // interations
```
aakoshh, line 66: What does this mean?

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Okay, I trust it will be done somewhere :+1:

aakoshh (approved):
Great :man_dancing:

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: I can remove that, but that was comments related to the `checks-effects-interactions` rule for Solidity to mitigate reentrancy attacks.

dnkolegov: and a misprint

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: I added that in `@dev` field and created an issue https://github.com/consensus-shipyard/ipc-solidity-actors/issues/227

src/interfaces/IGateway.sol
adlrocha: Already discussed sync. ACL already in place

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -29,32 +29,42 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
     function submitCheckpoint(
         BottomUpCheckpoint calldata checkpoint,
-        bytes32 membershipRootHash,
-        uint256 membershipWeight
+        CrossMsg[] calldata messages,
+        address[] calldata signatories,
+        bytes[] calldata signatures
     ) external {
+        // validations
+        if (checkpoint.blockHeight <= s.lastBottomUpCheckpointHeight) {
```
adlrocha, line 47: @dnkolegov, we are still not checking if the checkpoint committed for execution is the one for the next immediate height, right? Or am I missing something?

adlrocha (approved):
LGTM. I am still missing a check to ensure that we only commit the immediate next checkpoint next. As we are working with full-fat checkpoints first let's have this check and then improve once we have the low-fat approach.

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Sorry, I missed that one. It should be fixed now.

binding/src/subnet_actor_getter_facet.rs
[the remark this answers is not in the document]
dnkolegov: Thank you. Created a PR https://github.com/consensus-shipyard/ipc-solidity-actors/pull/231

PR #220, opened 2023-10-04
Clean up

cryptoAtwill:
Clean up unused code and also patching the unit tests so that new validator lifecycle works. Key updates:
- Removed unused fields
- Adding mocked testing contracts to decouple SubnetActor logic from Gateway tests.


src/SubnetActorDiamond.sol
```diff
@@ -51,10 +50,12 @@ contract SubnetActorDiamond {
         s.consensus = params.consensus;
         s.minActivationCollateral = params.minActivationCollateral;
         s.minValidators = params.minValidators;
-        s.topDownCheckPeriod = params.topDownCheckPeriod;
         s.bottomUpCheckPeriod = params.bottomUpCheckPeriod;
         s.status = Status.Instantiated;
         s.currentSubnetHash = s.parentId.createSubnetId(address(this)).toHash();
+
+        // 100 is the max supported now
```
adlrocha, line 57:
```suggestion
        //  We hardcode the current limit for active validators to 100 per Tendermint consensus
```

src/gateway/GatewayManagerFacet.sol
```diff
@@ -136,7 +123,7 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         // commit top-down message.
         LibGateway.commitTopDownMsg(crossMsg);
 
-        LibGateway.distributeRewards(subnetId.getActor(), s.crossMsgFee);
+        // Comment off first: LibGateway.distributeRewards(subnetId.getActor(), s.crossMsgFee);
```
adlrocha, line 126: I think we can remove this one, I feel rewards will work completely differently now with Fendermint (we can always come back to the previous release of the contracts to check this logic).

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -140,7 +140,10 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
         }
 
         if (shouldDistributeRewards) {
```
adlrocha, line 142: Let's probably remove all reward-related code here.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -86,9 +86,7 @@ contract GatewayRouterFacet is GatewayActorModifiers {
 
         _applyMessages(commit.subnetID, messages);
 
-        uint256 fee = 0;
-
-        LibGateway.distributeRewards(msg.sender, fee);
+        // TODO: distribute reward
```
adlrocha, line 89:
```suggestion
        // TODO: distribute rewards to validators for their service executing a checkpoint
```

adlrocha (commented):
Overall, the changes to the code make sense to me. I am wondering if we should still remove something from this data structures, I don't think all of them are currently used: https://github.com/consensus-shipyard/ipc-solidity-actors/blob/clean-up/src/structs/Validator.sol

Also, I rather have @dnkolegov have the final saying on this one, and for you two to agree on the strategy to recover the level of test coverage we had before. Thank you both!

cryptoAtwill:
@dnkolegov There are 8 tests commented off in gateway, we will add them back again. For SubnetActor, the `join` and other relevant processes are completely rewritten, the old tests will be eventually removed.

dnkolegov:
> @dnkolegov There are 8 tests commented off in gateway, we will add them back again. For SubnetActor, the `join` and other relevant processes are completely rewritten, and the old tests will be eventually removed.

If we removed unused tests and you are going to recover those 8 tests when it is possible I do not see problems.

dnkolegov (approved):

adlrocha (approved):

PR #223, opened 2023-10-04
IPC-221: Bootstrap subnets

### LINKED ISSUE #221: Remove `Status` from SubnetActor and subnet bootstrapping
## Background
Keeping `Status` in both the SubnetActor and the Gateway can lead to them having inconsistent states. `Status` in the SubnetActor was initially implemented to allow users to have custom requirements for their subnet, but once a subnet has been bootstrapped and activated, having a different `Status` from that in the gateway may be consume. 

We want subnet operators to be able to implement custom requirements to bootstrap their subnet. The current reference implementation only enforces two requirements: that the minimum threshold for validators have been reached, and that there is a minimum number of validators. 

On the other hand, the gateway enforces a minimum requirement of collateral for subnets to be able to perform cross-net operations (like submitting checkpoints, or sending cross-net messages).  

## Proposal
- Remove `Status` from the subnet actor. 
- Include bootstrapping logic to the subnet actor so an event is emitted when the minimum requirements for the subnet have been fulfilled. This event will be picked up by Fendermint to populate the genesis of the subnet and bootstrap the infrastructure for the child. 
- Keep the gateway as the "gatekeeper" of IPC through the `Status` of subnets. 
- Add `initialValidators` to the constructor of the gateway, so child subnets can already have their initial validators populated according to the ones picked up from the event from their parent.

adlrocha:
Closes #221.

Note: A few things that need to be addressed before merging.
- What data we want to include in `SubnetBootstrapped` to be picked up by Fendermint to populate the genesis and deploy the gateway. 
- A set of unit tests for the new functionality.
- Add initial validators to the gateway constructor parameters, so we can bootstrap child subnets with Fendermint.


src/lib/LibStaking.sol
```diff
@@ -360,6 +360,21 @@ library LibStaking {
...
+    }
+
+    /// @notice Checks if the validator is a waiting validator
+    function isWaitingValidator(address validator) internal view returns (bool) {
+        SubnetActorStorage storage s = LibSubnetActorStorage.appStorage();
+        return s.validatorSet.waitingValidators.contains(validator);
+    }
 
     /// @notice Checks if the validator is an active validator
     function isActiveValidator(address validator) internal view returns (bool) {
```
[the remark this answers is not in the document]
adlrocha: Yes, it was the result of my merge conflict fixing and forgot to push. Already fixed.

src/lib/LibStaking.sol
```diff
@@ -454,4 +469,26 @@ library LibStaking {
...
+
+    /// @notice Joins new validators when the subnet is not yet activated.
+    function initialJoin(address validator, uint256 amount) internal {
+        SubnetActorStorage storage s = LibSubnetActorStorage.appStorage();
+
+        // check if the validator is already in the active validators?
+
+        // confirm validators deposit immediately
+        s.validatorSet.confirmDeposit(validator, amount);
+        s.validatorSet.recordDeposit(validator, amount);
```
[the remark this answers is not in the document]
adlrocha: fixed.

src/lib/LibStaking.sol
```diff
@@ -360,6 +360,21 @@ library LibStaking {
 
     event ConfigurantionNumberConfirmed(uint64 number);
     event CollateralClaimed(address validator, uint256 amount);
+    // TODO: Include the list of initial validators as part of the event
+    // so Fendermint can directly pick it up when bootstrapping the infra?
+    event SubnetBootstrapped();
+
+    /// @notice Checks if the validator is an active validator
+    function isActiveValidator(address validator) internal view returns (bool) {
```
dnkolegov, line 368: https://github.com/consensus-shipyard/ipc-solidity-actors/blob/37807916b9f3b3bc646ed3004ab653688313e2f3/src/lib/LibStaking.sol#L194

Is it the same?

src/lib/LibStaking.sol
```diff
@@ -360,6 +360,21 @@ library LibStaking {
...
+        SubnetActorStorage storage s = LibSubnetActorStorage.appStorage();
+        return s.validatorSet.activeValidators.contains(validator);
+    }
+
+    /// @notice Checks if the validator is a waiting validator
+    function isWaitingValidator(address validator) internal view returns (bool) {
```
dnkolegov, line 374: Should we add this to the LibValidatorSet?

dnkolegov (approved):
Just a minor question about `active` and `waiting` validators.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -72,8 +68,15 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
 
-        LibStaking.setValidatorMetadata(msg.sender, metadata);
-        LibStaking.deposit(msg.sender, msg.value);
+        // if the subnet has not been bootstrapped, join directly
+        // without delays, and collect collateral to register
+        // in the gateway
+        if (!s.bootstrapped) {
+            LibStaking.initialJoin(msg.sender, msg.value);
+        } else {
+            LibStaking.setValidatorMetadata(msg.sender, metadata);
```
aakoshh, line 77: Why doesn't the metadata get set if the subnet is already bootstrapped?

src/lib/LibStaking.sol
adlrocha: Yes, it was fixed but I didn't push the changes sorry.

src/subnet/SubnetActorManagerFacet.sol
adlrocha: It was a bug from my side. I did already fix it locally but I didn't push it yet, sorry. Should be fixed now :)

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -72,8 +68,15 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        // if the subnet has not been bootstrapped, join directly
+        // without delays, and collect collateral to register
+        // in the gateway
         LibStaking.setValidatorMetadata(msg.sender, metadata);
-        LibStaking.deposit(msg.sender, msg.value);
+        if (!s.bootstrapped) {
+            LibStaking.initialJoin(msg.sender, msg.value);
+        } else {
+            LibStaking.deposit(msg.sender, msg.value);
+        }
```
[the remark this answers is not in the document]
adlrocha: Good point. Fixed in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/223/commits/a0cf1847fb5397764068911c6ba421f32901e1c1

src/lib/LibStaking.sol
```diff
@@ -360,6 +360,9 @@ library LibStaking {
 
     event ConfigurantionNumberConfirmed(uint64 number);
     event CollateralClaimed(address validator, uint256 amount);
+    // TODO: Include the list of initial validators as part of the event
+    // so Fendermint can directly pick it up when bootstrapping the infra?
```
aakoshh, line 364: It is essential to agree on what is the cutoff for generating the child genesis. I reckon it can be defined as the state after the execution of the block in which this event has been emitted. That way you can keep flexible what you put into the genesis, not just these validators. For example there might be joiners A, B, C an D and the criteria was met at C, but by looking at the final state we'd see D also joined together with the others, and you can also get other stuff needed that would not fit into this event.

aakoshh: OTOH I see that in my example D would already go into the change request queue. But then we can just query the active validators after the execution of this block to see who they are, right?

adlrocha: What I was envisioning is for the contract to handle this and have the "source of the truth": so if A, B, C, and D join in the same block, and their messages are executed in order being C the one that fulfills the requirements to bootstrap, then the contract will emit the event when C joins and those are the initial validators considered for config number C. When D joins, the new configuration will be assigned `configNum = 1` and will have to wait for the next checkpoint for Fendermint to apply the config change. 

I could maybe kick-off the next PR that handles this and discuss there?

adlrocha: > OTOH I see that in my example D would already go into the change request queue. But then we can just query the active validators after the execution of this block to see who they are, right?

 ^this! D will get into the queue already, and it would handle as any other update (even if it made it in the same block). The event is the cutoff for deterministic genesis generation

aakoshh (approved):
LGTM :+1:

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -105,8 +101,7 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
             revert NotAllValidatorsHaveLeft();
         }
 
-        s.status = Status.Killed;
-
+        s.killed = true;
```
aakoshh, line 104: Who can call `kill`?

aakoshh: Thanks for explaining. I don't see this check [here](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/f6e7ae686a67ed1e1e1f8002a81c1f4ca6941997/src/subnet/SubnetActorManagerFacet.sol#L100C9-L102C10) but as always I trust it exists somewhere. 

Just looking at the one that's there, though, the one that checks that the number of validators is 0, one has to wonder how the parent will ever have confirmation that there are 0 validators left. I mean in theory this is doable: current validators withdraw their collateral and produce a checkpoint in which they tell CometBFT that in the next block every validator has 0 power. I'm not sure what CometBFT will say about that, but if it means the end of the road, then there won't be a next block where the current validators can submit their signatures, no next block for the relayer to pick up the event, etc. 

So, the validators would be able to gossip their signatures directly between each other until one of them can produce a checkpoint, it would not be with the mechanisms we are developing today.

adlrocha: I see. You are right, we should completely re-define how we kill subnets with the new method. Actually, let me add it as something to address in the next milestone (I don't think we have space in this one, but this is critical to have the full story fleshed out)

aakoshh: We could possibly tell CometBFT to keep producing blocks with the current validator set, and then just fail to ever send a newer checkpoint.

PR #225, opened 2023-10-05
store validator changes

cryptoAtwill:
Store validator change set from parent into child.


src/lib/LibGatewayActorStorage.sol
```diff
@@ -74,6 +74,8 @@ struct GatewayActorStorage {
     uint64 bottomUpCheckPeriod;
     /// @notice whether the contract is initialized
     bool initialized;
+    /// Validator tracking changes from parent in child subnet
```
aakoshh, line 77:
```suggestion
    /// Tracking validator changes from parent in child subnet
```

src/lib/LibStaking.sol
```diff
@@ -52,6 +48,19 @@ library LibStakingChangeSet {
         });
     }
 
+    /// @notice Perform upsert operation to the deposit changes
+    function recordChange(
+        StakingChangeSet storage changes,
```
aakoshh, line 53: I got confused by the "ChangeSet" naming, I thought it suddenly means there are multiple changes are bundled together, which was something I thought we'd do if we submitted deltas. 

Perhaps we could rename `StakingChangeSet` to `StakingChangeLog`. After all it's not a _set_ but an ordered list of changes.

test/GatewayDiamond.t.sol
```diff
@@ -1206,10 +1204,14 @@ contract GatewayDiamondDeploymentTest is StdInvariant, Test {
...
-        vm.expectRevert(ValidatorsAndWeightsLengthMismatch.selector);
 
         ParentFinality memory finality = ParentFinality({height: block.number, blockHash: bytes32(0)});
-        gwRouter.commitParentFinality(finality, n, validators, weights);
+        gwRouter.commitParentFinality(finality);
+
+        vm.startPrank(FilAddress.SYSTEM_ACTOR);
+        vm.expectRevert(ValidatorsAndWeightsLengthMismatch.selector);
+
+        gwRouter.updateMembership(n, validators, weights);
```
adlrocha, line 1214: This applies to all the tests, I see that here we are just calling the functions. Would it be worth making some sanity-checks to ensure that the state changes are the right ones?

src/lib/LibStaking.sol
```diff
@@ -455,3 +464,38 @@ library LibStaking {
         emit ConfigurantionNumberConfirmed(configurationNumber);
     }
 }
+
+/// The library for tracking waiting and active validators, should be used in gateway to track the
+/// validator change set and apply them directly.
```
aakoshh, line 469:
```suggestion
/// The library for tracking validator changes coming from the parent.
/// Should be used in the child gateway to store changes until they can be applied.
```

src/lib/LibStaking.sol
```diff
@@ -455,3 +464,38 @@ library LibStaking {
...
+            op: changeRequest.change.op,
+            amount: changeRequest.change.amount
+        });
+
+        if (configurationNumber != changeRequest.configurationNumber) {
+            revert InvalidConfigurationNumber();
+        }
+    }
+
+    function storeChanges(
```
adlrocha, line 486: Strong opinion weakly held, but I am usually not a fan of single letter diffs between functions :) Maybe is my disastrous personality haha
```suggestion
    function batchStoreChange(
```

src/structs/Subnet.sol
```diff
@@ -109,6 +115,12 @@ struct ValidatorSet {
     MaxPQ waitingValidators;
 }
 
+/// Tracks the parent validator changes and apply them in the child
+struct ParentValidatorsTracker {
+    ValidatorSet validators;
```
adlrocha, line 120: What is this field used for? (I couldn't find where is used in the code)

src/structs/Subnet.sol
```diff
@@ -39,6 +39,12 @@ struct StakingChange {
     address validator;
 }
 
+/// The change associated with its corresponding configuration number
+struct StakingChangeRequest {
+    StakingChange change;
```
aakoshh, line 44: How is the public key of the validator going to be propagated to the child subnet?

src/structs/Subnet.sol
cryptoAtwill: it will be used when the changes are applied to track active validators. I can remove them first.

src/structs/Subnet.sol
cryptoAtwill: I will tackle it in another PR.

test/GatewayDiamond.t.sol
cryptoAtwill: This function will be deprecated with new validators. It was the same implementation as before.

src/structs/Subnet.sol
cryptoAtwill: Here it is: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/230

adlrocha (approved):

PR #228, opened 2023-10-05
IPC-221: include genesisValidators and bootstrapped event

adlrocha:
Follow-up to #223.

This PR: 
- Stores `genesisValidators` in the subnet actor. This information will be used by Fendermint to deterministically create the genesis for the child subnets in order to bootstrap it.
-  Emits the genesis validators in an event when the subnet is bootstrapped.
- Includes `genesisValidators` as a parameter to the gateway constructor so we can set in construction the genesis membership for the subnet.


src/subnet/SubnetActorGetterFacet.sol
```diff
@@ -47,6 +47,10 @@ contract SubnetActorGetterFacet {
         return (s.changeSet.nextConfigurationNumber, s.changeSet.startConfigurationNumber);
     }
 
+    function genesisValidators() external view returns (GenesisValidator[] memory) {
```
aakoshh, line 50: This certainly makes it easier to query rather than having to find a height at which the active validator set is correct.

src/GatewayDiamond.sol
```diff
@@ -22,6 +25,7 @@ contract GatewayDiamond {
         uint256 minCollateral;
         uint256 msgFee;
         uint8 majorityPercentage;
+        Validator[] genesisValidators;
```
aakoshh, line 28: Why isn't this `GenesisValidator[]` ?

src/structs/Subnet.sol
```diff
@@ -85,6 +85,15 @@ struct Validator {
     bytes metadata;
 }
 
+/// Struct that holds all the information required for a genesis vallidator,
+/// including its address and genesis collateral and metadata.
+/// This can be used by Fendermint to bootstrap a new child subnet.
+struct GenesisValidator {
+    address addr;
+    uint256 genesisCollaterall;
+    bytes metadata;
```
aakoshh, line 94: Shouldn't `metadata` just be added to the `Validator`? Yes, we discussed in https://github.com/consensus-shipyard/fendermint/issues/304 that `metadata` will be part of the changes, but surely it will also need to be stored somewhere, as it will have to be returned during the next validator update cycle.

aakoshh: I mean, technicall for https://github.com/consensus-shipyard/fendermint/issues/306 it's enough for `GenesisValidator` to have it, but in https://github.com/consensus-shipyard/fendermint/issues/301 I reckon if we can't store this metadata in the contract, then in the scenarios described in https://github.com/consensus-shipyard/fendermint/issues/305 I will have trouble figuring out which PublicKey in CometBFT belongs to which address unless we demand that the address is a hash of the public key.

src/structs/Validator.sol
```diff
@@ -3,6 +3,11 @@ pragma solidity 0.8.19;
 
 import {FvmAddress} from "../structs/FvmAddress.sol";
 
+// These structs use `FvmAddress` instead of plain `address` because they need to
+// trespass subnet boundaries. As a rule of thumb, we use `address` for everything
+// that is treated exclusively in Solidity, and `FvmAddress` for things that involve
+// the FVM or may be impacted by having heterogeneous runtimes.
```
aakoshh, line 9: So we discussed that to be a validator implies one had to invoke a Solidity contract to join a subnet, which can technically [only be done](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/message/src/signed.rs#L109) from an f410 address, and therefore it would simplify other operations such as genesis file construction if we did not have to entertain the idea of different kind of addresses here. 

For example if we know the public key of validators, and agree that there is a 1-to-1 correspondence between the public key and the validator address, we can trivially produce the f410/ethereum address by hashing, which keeps the UX simpler.

src/GatewayDiamond.sol
```diff
@@ -48,6 +52,21 @@ contract GatewayDiamond {
         s.crossMsgFee = params.msgFee;
         s.majorityPercentage = params.majorityPercentage;
         s.bottomUpCheckpointRetentionHeight = 1;
+
+        // set initial validators and update membership
+        FvmAddress[] memory validators = new FvmAddress[](params.genesisValidators.length);
+        uint256[] memory weights = new uint256[](params.genesisValidators.length);
+        for (uint256 i = 0; i < params.genesisValidators.length;) {
```
dnkolegov, line 59: The length should be saved outside of the loop on the line 56

src/lib/LibStaking.sol
```diff
@@ -410,8 +410,30 @@ library LibStaking {
             // without delays, and collect collateral to register
             // in the gateway
             // confirm validators deposit immediately
+            s.validatorSet.recordDeposit(validator, amount);
             s.validatorSet.confirmDeposit(validator, amount);
 
+            // add to initial validators avoiding duplicates
+            bool alreadyValidator = false;
+            for (uint256 i = 0; i < s.genesisValidators.length; ) {
```
dnkolegov, line 418: The same regarding the length

src/lib/LibStaking.sol
```diff
@@ -410,8 +410,30 @@ library LibStaking {
...
+                if (s.genesisValidators[i].addr == validator) {
+                    alreadyValidator = true;
+                    break;
+                }
+                unchecked {
+                    i++;
```
dnkolegov, line 424: ++i

dnkolegov (approved):
LGTM. I added three minor comments related Solidity best practices

src/structs/Subnet.sol
```diff
@@ -85,6 +85,15 @@ struct Validator {
     bytes metadata;
 }
 
+/// Struct that holds all the information required for a genesis vallidator,
+/// including its address and genesis collateral and metadata.
+/// This can be used by Fendermint to bootstrap a new child subnet.
+struct GenesisValidator {
+    address addr;
+    uint256 genesisCollaterall;
```
aakoshh, line 93:
```suggestion
    uint256 collateral;
```

src/GatewayDiamond.sol
adlrocha: Fixed. But out of a curiosity, why do we need this? Does Solidity re-compute the length in every iteration of the loop?

src/lib/LibStaking.sol
adlrocha: fixed

src/lib/LibStaking.sol
adlrocha: fixed

src/GatewayDiamond.sol
dnkolegov: It does, as far a I know.

src/structs/Subnet.sol
adlrocha: fixed.

PR #229, opened 2023-10-05
Checkpoint execution

dnkolegov:
This PR implements a part of https://github.com/consensus-shipyard/fendermint/issues/295


src/gateway/GatewayRouterFacet.sol
```diff
@@ -73,17 +66,22 @@ contract GatewayRouterFacet is GatewayActorModifiers {
             }
         }
 
+        // CODE-REVIEW
```
adlrocha, line 69: In the previous design, checkpoints included a `fee` field that was increased in the child subnet whenever there was a new cross net message that released funds, as a payment for validators. These fees were burnt in the child, and send in the checkpoint with other cross-net messages. We could include that same field again, and let validators vote the fee they want to propagate up. For now, we can either assume that this fee exists (and even include it) or add a placeholder here and do nothing until we figure out the incentive system in the next milestone.

src/interfaces/IGateway.sol
```diff
@@ -18,13 +18,19 @@ interface IGateway {
     /// Release stake recovers some collateral of the subnet
     function releaseStake(uint256 amount) external;
 
+    // Release rewards to the validators
+    function releaseRewardsForValidators(uint256 amount) external;
```
adlrocha, line 22: Instead of having a different function to claim the stake released, and this one to claim rewards, would it be hard to merge both of them into `claim`. That way with a single function validators can claim all their outstanding funds in a subnet.

src/lib/LibStaking.sol
```diff
@@ -170,6 +170,18 @@ library LibValidatorSet {
         collateral = validators.validators[validator].confirmedCollateral;
     }
 
+    function listActiveValidators(ValidatorSet storage validators) internal view returns (address[] memory addresses) {
+        uint16 size = validators.activeValidators.getSize();
+        addresses = new address[](size);
+        for (uint16 i = 1; i <= size; ) {
```
adlrocha, line 176: It feels weird to me to start from 1. Any strong reason for this?

src/lib/LibStaking.sol
```diff
@@ -170,6 +170,18 @@ library LibValidatorSet {
         collateral = validators.validators[validator].confirmedCollateral;
     }
 
+    function listActiveValidators(ValidatorSet storage validators) internal view returns (address[] memory addresses) {
```
adlrocha, line 173: IIRC, I use a similar function here https://github.com/consensus-shipyard/ipc-solidity-actors/pull/228 to iterate through active validators. We should maybe de-duplicate them

src/lib/priority/LibMinPQ.sol
```diff
@@ -18,6 +18,10 @@ library LibMinPQ {
         return self.inner.size;
     }
 
+    function getAddress(MinPQ storage self, uint16 i) internal view returns (address) {
```
adlrocha, line 21: Hmm, now that I think of it, this function may also come handy for my PR. We should maybe consider merging this one first and then me rebasing to this one

adlrocha (commented):
I'll probably need another pass before we merge, but overall it's looking good. So far the execution of messages in `commitBottomUpCheckpoint` for the gateway stays unchanged, right?

src/lib/LibStaking.sol
dnkolegov: That's @cryptoAtwill design decision.
See
- https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1696327777276459?thread_ts=1696327054.120109&cid=C04JR5R1UL8
- https://github.com/consensus-shipyard/ipc-solidity-actors/blob/99b4acbfce2967fac6adbfc10ebbe9884c42d918/src/lib/priority/LibPQ.sol#L14

src/lib/LibStaking.sol
dnkolegov: I could not find it. Could you point me to it? Thank you.

src/lib/priority/LibMinPQ.sol
dnkolegov: added a todo

src/lib/LibStaking.sol
dnkolegov: added a todo

src/interfaces/IGateway.sol
dnkolegov: Improved, but not ideally. Please check.

src/gateway/GatewayManagerFacet.sol
```diff
@@ -65,8 +65,9 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         }
     }
 
-    /// @notice release collateral for an existing subnet
-    function releaseStake(uint256 amount) external nonReentrant {
+    /// @notice release amount for an existing subnet
+    /// @dev it can be used to release the stake or reward of the validator
+    function releaseAmount(uint256 amount) external nonReentrant {
```
aakoshh, line 70: I'm admittedly a Solidity noob, but I don't really understand the implementation of this method. 

I added some questions to the body here:

```solidity
    function releaseAmount(uint256 amount) external nonReentrant {
        if (amount == 0) {
            revert CannotReleaseZero();
        }

        // What does it mean to get a subnet of the sender? Is this running in the parent or the child?
        (bool registered, Subnet storage subnet) = LibGateway.getSubnet(msg.sender);

        if (!registered) {
            revert NotRegisteredSubnet();
        }
        // The function says this can release rewards. Do rewards end up in the subnet stake?
        // Is stake same as collateral? Is it for the whole subnet, or is it this validator's stake?
        if (subnet.stake < amount) {
            revert NotEnoughFundsToRelease();
        }
 
        // Is amount common to all validators, or is it specifically for the sender?
        subnet.stake -= amount;

        if (subnet.stake < s.minStake) {
            subnet.status = Status.Inactive;
        }

        // Does this mean we are sending the amount to the subnet actor? 
        // What does that mean, why not send it to the validator?
        payable(subnet.id.getActor()).sendValue(amount);
    }
```

aakoshh: Maybe this method is missing a check like this?

https://github.com/consensus-shipyard/ipc-solidity-actors/pull/229/files#diff-595ecb0d30017b2295ddc7f0716e5e0b525a44e04e314b80304622ef1a03d099R48

It would be helpful to mention in the method description all these constraints. I'm sorry but I'm usually utterly lost.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +41,69 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
         }
-
-        // slither-disable-next-line unused-return
-        (, Subnet storage subnet) = LibGateway.getSubnet(msg.sender);
+        (bool subnetExists, Subnet storage subnet) = LibGateway.getSubnet(msg.sender);
+        if (!subnetExists) {
+            revert SubnetNotFound();
+        }
+        if (checkpoint.subnetID.equals(subnet.id)) {
+            revert InvalidSubnet();
```
aakoshh, line 56: I don't understand, what should it equal to if not the sender subnet actor?

dnkolegov: Good catch! The correct check should be `checkpoint.subnetID.equals(subnet.id)`. Thanks.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +41,69 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
-        for (uint256 i = 0; i < crossMsgLength; ) {
-            totalValue += messages[i].message.value;
+    /// @notice reward the relayers for processing checkpoint at height `h`.
+    /// @dev the relayer is added to the list of all relayers for this checkpoint and rewarded at the time the next checkpoint arrives,
+    /// so that we can distribute a fixed reward pot among them, rather than reward them at the time of submission.
+    function rewardRelayers(uint64 h, uint256 fee) external {
+        address[] memory relayers = s.rewardedRelayers[h].values();
+        uint256 n = relayers.length;
+        for (uint256 i = 0; i < n; ) {
+            LibGateway.distributeRewards(relayers[i], fee);
```
aakoshh, line 102: Thank you for including this! 

I have two questions:
* I understand it's `external` because the subnet actor has to call it, but then it seems to be missing the check [like this](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/229/files#diff-595ecb0d30017b2295ddc7f0716e5e0b525a44e04e314b80304622ef1a03d099R48) to make sure the relayers themselves cannot call this again and again.
* By "distribute a fixed reward pot among them" I meant that we divide it among them, so it would be `LibGateway.distributeRewards(relayers[i], fee / n);`

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +41,69 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
-            checkpoint.subnetID = s.networkName;
-            checkpoint.blockHeight = nextCheckEpoch;
+            _applyMessages(checkpoint.subnetID, messages);
         }
 
-        CrossMsg[] memory messages = s.bottomUpMessages[commit.blockHeight];
+        // The relayer is added to the list of all relayers for this checkpoint
+        // to be rewarded later.
+        // slither-disable-next-line unused-return
+        s.rewardedRelayers[checkpoint.blockHeight].add(msg.sender);
```
aakoshh, line 92: There could potentially be some kind of check that rewards haven't been paid out yet. It would be irrational to send more relays after even the possibility of a reward is gone, but if you wanted to see statistics of how many relayers there were, it would be better if this was sealed at the point when the next checkpoint arrived (or whatever triggered the payout).

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +41,69 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+            for (uint256 i = 0; i < crossMsgLength; ) {
+                totalValue += messages[i].message.value;
+                unchecked {
+                    ++i;
+                }
+            }
+
+            // CODE-REVIEW
+            // before we do this here totalValue += commit.fee + checkpoint.fee
+            // where and how are we going to use fees?
```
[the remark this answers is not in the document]
dnkolegov: Created an issue to address it separately in a new story - https://github.com/consensus-shipyard/ipc-solidity-actors/issues/232

src/lib/LibStaking.sol
aakoshh: I think it may be because [contains checks](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/99b4acbfce2967fac6adbfc10ebbe9884c42d918/src/lib/priority/LibPQ.sol#L38) that the position of an address is not 0, which IIUC is the default value you'd get if you look up a value in a mapping which doesn't have it.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -115,11 +115,38 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+
+        uint256 validatorsLength = validators.length;
+        if (validatorsLength == 0) {
+            revert NoValidatorsInSubnet();
+        }
+        if (amount < validatorsLength) {
+            revert NotEnoughBalanceForRewards();
+        }
+
+        uint256 rewardAmount = amount / validatorsLength;
```
aakoshh, line 140: Shouldn't it be proportional to their weight?

src/lib/LibStaking.sol
```diff
@@ -453,6 +468,22 @@ library LibStaking {
         emit CollateralClaimed(validator, amount);
     }
 
+    /// @notice method that allows a validator to withdraw it's accumulated rewards using pull-based transfer
+    function claimReward(address validator) external {
```
aakoshh, line 472: `claimRewards` is supposed to be called on the parent, right? You go to a subnet you are validator in, and want to claim your rewards. This is not a reward for all the _other_ validators on the parent subnet, right? 

What are the rewards for exactly? I mean, child subnet validators get fees in the child, and parent validators get fees in the parent for executing the cross messages. Where is this reward for the child validators _in_ the parent subnet coming from?

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -115,11 +115,38 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
         LibStaking.claimCollateral(msg.sender);
     }
 
+    /// @notice Validator claims their released reward
+    function claimReward() external {
+        LibStaking.claimReward(msg.sender);
+    }
+
+    /// @notice method that distributes the rewards for the subnet to validators.
+    function rewardValidators(uint256 amount) external onlyGateway {
```
aakoshh, line 129: Where is this called from?

src/lib/LibStaking.sol
```diff
@@ -453,6 +468,22 @@ library LibStaking {
...
+        SubnetActorStorage storage s = LibSubnetActorStorage.appStorage();
+        uint256 amount = s.accumulatedRewards[validator];
+
+        if (amount == 0) {
+            revert NoRewardToWithdraw();
+        }
+
+        s.accumulatedRewards[validator] = 0;
+
+        IGateway(s.ipcGatewayAddr).releaseAmount(amount);
```
aakoshh, line 482: I might be misunderstanding something, but I don't think this method should be called. 

As I understand, the gateway locks funds on behalf of the subnet actor, and `releaseAmount` sends some of that back to the subnet actor, who then forwards it to the validator. 

However the `releaseAmount` function will also lower the overall stake (collateral) behind the subnet, which suggests it should only be called when collateral is released, not rewards. 

Could you also confirm when the rewards are locked up in the gateway?

src/gateway/GatewayRouterFacet.sol
dnkolegov: When does`commitBottomUpCheckpoint` stop accepting checkpoints for some height?

src/gateway/GatewayRouterFacet.sol
dnkolegov: Here we call a method on the SubnetActor and that method [divides](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/fbba77dd83c1482cbd9b8278fd49740db7e9ea6c/src/subnet/SubnetActorManagerFacet.sol#L129) the fee among all relayers.

Only the gateway is allowed to call it.

src/lib/LibStaking.sol
dnkolegov: A concrete validator calls `claimReward` to get its reward for submitting checkpoints. This is `pull`-pattern to get rewards assigned in `commitBottomUpCheckpoint`

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: It should. Fixed

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Do you have a link? Sorry, I don't know which file it's in and I marked all of them as viewed, so it's needle in the haystack.

dnkolegov: https://github.com/consensus-shipyard/ipc-solidity-actors/blob/15ffc10cd509468cc5a7a93bbaae9e6b460499c6/src/lib/LibGateway.sol#L272

src/lib/LibStaking.sol
aakoshh: Thanks, so it's given in exchange for signing the checkpoint. 

What's not entirely clear to me is who should be paying this reward? What rewards can validators expect if there are no cross messages?

src/gateway/GatewayRouterFacet.sol
aakoshh: I thought a natural place is when the next checkpoint arrives. 
1. Checkpoint N submitted by relayer 1 -> add relayer 1
2. Checkpoint N executed
3. Checkpoint N submitted again by relayer 2 -> add relayer 2
4. Checkpoint N+1 submitted by relayer 2 -> add relayer 1
5. Checkpoint N+1 executed -> reward relayer 1 and 2 for checkpoint N
6. Checkpoint N submitted by relayer 3 -> ignore, we already executed checkpoint N+1

src/gateway/GatewayRouterFacet.sol
aakoshh: I'm not sure I follow. You are pointing at a method called `rewardValidators`. This one is `rewardRelayers`. I interpret `fee` as the fixed reward pot, and then it distributes it among all the validators. 

I'm not sure it should be called `fee`, rather than `reward`. My idea is simple: validators offer e.g. 10 tokens for the relay of a checkpoint, in total. If one relayer sends it, they get 10. If two of them send it, they get 5 each. 

Is that happening somewhere? Is `fee` already divided by `relayers.length`? The comments don't say.

src/gateway/GatewayManagerFacet.sol
dnkolegov: The link points to the entire diff bot, not to the concrete method.

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Thanks.. Sorry, who calls `distributeRewards` ? :sweat_smile:

src/gateway/GatewayManagerFacet.sol
aakoshh: Sorry, here's what it's pointing at, indeed in a different method:
```
        // checkpoint is used to implement access control
        if (checkpoint.subnetID.getActor().normalize() != msg.sender) {
            revert InvalidCheckpointSource();
        }
```

My point is that `releaseAmount` should probably validate it's being called by the subnet actor.

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: - If GatewayActor's `rewardRelayers` is called then it [calls](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/15ffc10cd509468cc5a7a93bbaae9e6b460499c6/src/lib/LibGateway.sol#L272) `distributeRewards` for each validator.

- `distributeRewards` [calls](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/15ffc10cd509468cc5a7a93bbaae9e6b460499c6/src/lib/LibGateway.sol#L277) on the Subnetactor.
- `rewardValidators` can be [called](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/ca6849a26d06ea9809b2cfd2d7f2009d1c13df05/src/subnet/SubnetActorManagerFacet.sol#L129) by the gateway only. That function rewards a validator with an amount proportional to its power. That reward is recorded in the state.

- If later, a validator wants to get its rewards, it calls `claimReward()` from the Subnetactor, and that reward will be sent to it.
- When `claimReward` is called by the validator, it checks whether the validator has accumulated rewards in the state, and if it does, the actor [calls](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/fbba77dd83c1482cbd9b8278fd49740db7e9ea6c/src/lib/LibStaking.sol#L482) `releaseAmount` on the GatewayActor for accounting and send that amount to the validator.

src/gateway/GatewayManagerFacet.sol
dnkolegov: You are right. We use the same approach for access control there as in the checkpointing. It is there https://github.com/consensus-shipyard/ipc-solidity-actors/blob/fbba77dd83c1482cbd9b8278fd49740db7e9ea6c/src/gateway/GatewayManagerFacet.sol#L75

dnkolegov: > I added some questions to the body here:

The flow is described in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/229#discussion_r1350371510

> What does it mean to get a subnet of the sender? Is this running in the parent or the child?

Since `releaseAmount` is called by `SubnetActor`, `msg.sender` will be the subnet actor in the parent.

I reverted the implementation to use `releaseStake` and `releaseReward`, which we removed one week ago. 

Stake and collateral are used interchangeably in the code.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Renamed to `reward`.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,90 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
-        if (commit.subnetID.getActor().normalize() != msg.sender) {
+    /// @notice submit a verified checkpoint in the gateway to trigger side-effects and apply cross-messages.
+    /// Called from a subnet actor if the checkpoint is cryptographically valid.
+    function commitBottomUpCheckpoint(BottomUpCheckpoint calldata checkpoint, CrossMsg[] calldata messages) external {
+        // basic checks
+        // TODO: store checkpoint reward in the contract
+        uint256 checkpointReward = 10 gwei;
+
+        // checkpoint is used to implement access control
+        if (checkpoint.subnetID.getActor().normalize() != msg.sender) {
```
adlrocha, line 54: Why do you need to normalize here? Can't you directly check the actor with the `msg.sender`? They should both be Ethreum addresses.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,90 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
-    // TODO: Reimplement the function.
-    /// @dev This function must be reimplemented according to the current checkpoint protocol.
-    /// @notice submit a checkpoint in the gateway. Called from a subnet once the checkpoint is voted for and reaches majority
-    function commitChildCheck(BottomUpCheckpoint calldata commit) external {
-        if (commit.subnetID.getActor().normalize() != msg.sender) {
+    /// @notice submit a verified checkpoint in the gateway to trigger side-effects and apply cross-messages.
+    /// Called from a subnet actor if the checkpoint is cryptographically valid.
+    function commitBottomUpCheckpoint(BottomUpCheckpoint calldata checkpoint, CrossMsg[] calldata messages) external {
+        // basic checks
+        // TODO: store checkpoint reward in the contract
```
adlrocha, line 50: At this point, I would suggest keeping a placeholder for the reward mechanism and then figure out the amounts that we need. The key here is that we'll need to: 
* Have a way to collect fees from cross-net messages that then can be used to reward relayers. 
* Compute the payout that is paid to relayers for their job. 

I don't think there is an issue with this, I'll add it throughout the day so we can discuss the design.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,90 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
             revert InvalidCheckpointSource();
         }
-
-        // slither-disable-next-line unused-return
-        (, Subnet storage subnet) = LibGateway.getSubnet(msg.sender);
+        (bool subnetExists, Subnet storage subnet) = LibGateway.getSubnet(msg.sender);
+        if (!subnetExists) {
+            revert SubnetNotFound();
+        }
+        if (!checkpoint.subnetID.equals(subnet.id)) {
```
adlrocha, line 61: While I think this check is redundant, it may be useful to avoid potential replay attacks from different parent (although I feel it would be impossible, as messages already have replay attack protection through a `chainID`)

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,90 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
-        (bool checkpointExists, uint64 nextCheckEpoch, BottomUpCheckpoint storage checkpoint) = LibGateway
-            .getCurrentBottomUpCheckpoint();
+        if (checkpoint.blockHeight == s.lastBottomUpExecutedCheckpointHeight + s.bottomUpCheckPeriod) {
+            // If the submission for the next expected height then this is a new checkpoint which must be executed,
+            // the relayer must be stored, last executed subnet updated.
+
+            uint256 totalValue = 0;
+            uint256 crossMsgLength = messages.length;
+            for (uint256 i = 0; i < crossMsgLength; ) {
+                // CODE-REVIEW
```
adlrocha, line 75: As mentioned on my message above, I'll try to have an issue that explains how we can handle fees, but ideally we should leave it out of scope for this milestone.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,90 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+    function rewardRelayers(address actor, uint64 height, uint256 reward) internal {
+        if (reward == 0) {
+            return;
+        }
+        address[] memory relayers = s.rewardedRelayers[height].values();
+        if (relayers.length == 0) {
+            return;
+        }
+        // slither-disable-next-line unused-return
+        Address.functionCall(actor.normalize(), abi.encodeCall(ISubnetActor.rewardRelayers, (relayers, reward)));
```
adlrocha, line 129: Same question, why do you need to `normalize` if this is expected to be already an Ethereum address?

src/interfaces/IGateway.sol
```diff
@@ -8,23 +8,29 @@ import {FvmAddress} from "../structs/FvmAddress.sol";
...
-    /// AddStake adds stake to the collateral of a subnet.
+    /// @notice AddStake adds stake to the collateral of a subnet.
     function addStake() external payable;
 
-    /// Release stake recovers some collateral of the subnet
+    /// @notice Release stake recovers some collateral of the subnet
     function releaseStake(uint256 amount) external;
 
-    // Kill propagates the kill signal from a subnet actor to unregister it from th
+    /// @notice // Release reward to the subnet actor
```
adlrocha, line 21:
```suggestion
    /// @notice Release reward to the subnet actor
```

src/gateway/GatewayRouterFacet.sol
adlrocha: Also, I don't think we should handle relayer rewards in the gateway but in the subnet actor. Why do you need this function here that then calls the subnet actor instead of handling rewards directly from the subnet actor?

src/lib/LibStaking.sol
adlrocha: Agree here with @aakoshh. Validators are not rewarded for relaying bottom-up checkpoints. Whatever reward they should get for a cross-net interaction must be paid in their subnet already (it also make thinks simpler). So I would remove this one in favor of the relayer reward mechanism included above.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -141,11 +146,35 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
     }
 
-    /// @notice Valdiator claims their released collateral
+    /// @notice Validator claims their released collateral
     function claim() external nonReentrant {
         LibStaking.claimCollateral(msg.sender);
     }
 
+    /// @notice Validator claims their released reward
+    function claimReward() external {
```
adlrocha, line 155: I don't think we need this function per the comment above.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -141,11 +146,35 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        LibStaking.claimReward(msg.sender);
+    }
+
+    /// @notice method that distributes the rewards for the subnet to relayers.
+    function rewardRelayers(address[] memory relayers, uint256 reward) external onlyGateway {
```
adlrocha, line 160: Nice one, so we determine the total reward and (for now) distribute equally among all the validators for the checkpoint.

adlrocha (commented):
New round of reviews. My main comment is that I feel that the reward payouts to relayers should be entirely in the subnet actor instead of requiring that dance of the gateway notifying the rewards to the subnet actor, and the subnet actor paying out.  I think the subnet actor should be the one doing the bookkeeping of rewards, and the payout. 

I would also suggest moving into the execution of messages as soon as possible to get a better understanding of the end-to-end.

Let me know if there is any outstanding question/comment that requires my attention, if not let's address this round and I think we are close to being able to merge. 

Additional note, I am a bit worried that we are neglecting the implementation of basic unit tests for the last few PRs :( We may regret this in the future

src/gateway/GatewayRouterFacet.sol
dnkolegov: Ok. Fixed

src/interfaces/IGateway.sol
dnkolegov: Fixed.

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Removed

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Correct.

src/lib/LibStaking.sol
adlrocha: You can remove the TODO, I ended up not implementing mine, sorry.

src/lib/priority/LibMinPQ.sol
adlrocha: You can remove the TODO, I ended up not implementing it myself.

dnkolegov: ok

src/lib/LibStaking.sol
dnkolegov: ok

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -43,11 +43,12 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         address[] calldata signatories,
         bytes[] calldata signatures
     ) external {
-        // validations
+        // 1. Validate the checkpoint.
+
         if (!LibStaking.isActiveValidator(msg.sender)) {
             revert NotValidator();
         }
```
aakoshh, line 50: Not needed any more.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -59,14 +60,33 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
             revert InvalidCheckpointMessagesHash();
         }
         bytes32 checkpointHash = keccak256(abi.encode(checkpoint));
+        // validate signatures and quorum threshold, revert if validation fails
         validateActiveQuorumSignatures({signatories: signatories, hash: checkpointHash, signatures: signatures});
```
aakoshh, line 64: This is where it can only hope to succeed if `checkpoint.blockHeight == s.lastBottomUpCheckpointHeight + s.bottomUpCheckPeriod` but not if ` checkpoint.blockHeight == s.lastBottomUpCheckpointHeight`

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -59,14 +60,33 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        if (checkpoint.blockHeight == s.lastBottomUpCheckpointHeight + s.bottomUpCheckPeriod) {
+            // If the checkpoint height is the next expected height then this is a new checkpoint which must be executed
+            // in the Gateway Actor, the checkpoint and the relayer must be stored, last bottom-up checkpoint updated.
+
+            s.committedCheckpoints[checkpoint.blockHeight] = checkpoint;
+
+            // slither-disable-next-line unused-return
+            s.rewardedRelayers[checkpoint.blockHeight].add(msg.sender);
+
+            s.lastBottomUpCheckpointHeight = checkpoint.blockHeight;
```
aakoshh, line 77: Why do you set this just before it [will be checked](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/229/commits/c346939c731e30263c2b2a673033ba7dbddfeb2f#diff-595ecb0d30017b2295ddc7f0716e5e0b525a44e04e314b80304622ef1a03d099R68) in the gateway that this is _not_ the case?

I thought it was the gateway's responsibility to [set this field](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/229/commits/c346939c731e30263c2b2a673033ba7dbddfeb2f#diff-595ecb0d30017b2295ddc7f0716e5e0b525a44e04e314b80304622ef1a03d099R98).

dnkolegov: These are two different fields. gatewayActor will do the same but for `lastBottomUpExecutedCheckpointHeight`

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: But then any relayer can send checkpoint with the same height and get rewards

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Ah, thanks, sorry, I'm going blind

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,64 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
-    /// @dev This function must be reimplemented according to the current checkpoint protocol.
-    /// @notice submit a checkpoint in the gateway. Called from a subnet once the checkpoint is voted for and reaches majority
-    function commitChildCheck(BottomUpCheckpoint calldata commit) external {
-        if (commit.subnetID.getActor().normalize() != msg.sender) {
+    /// @notice submit a verified checkpoint in the gateway to trigger side-effects and apply cross-messages.
+    /// @dev this method is called by the corresponding subnet actor.
+    /// Called from a subnet actor if the checkpoint is cryptographically valid.
+    function commitBottomUpCheckpoint(BottomUpCheckpoint calldata checkpoint, CrossMsg[] calldata messages) external {
+        // TODO: store checkpoint reward in the contract and implement the proper reward mechanism later
+        uint256 checkpointReward = 10 gwei;
```
aakoshh, line 51: Maybe we can let the subnet actor decide how much it wants to pay for relayers?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,64 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+            // reward relayers in the subnet for committing the previous checkpoint
+            // slither-disable-next-line unused-return
+            Address.functionCall(
+                msg.sender,
+                abi.encodeCall(ISubnetActor.rewardRelayers, (s.lastBottomUpExecutedCheckpointHeight, checkpointReward))
+            );
+
+            // seal the checkpoint for the height `lastBottomUpExecutedCheckpointHeight`
+            s.lastBottomUpExecutedCheckpointHeight = checkpoint.blockHeight;
+        } else if (checkpoint.blockHeight == s.lastBottomUpExecutedCheckpointHeight) {
```
aakoshh, line 99: When does the subnet actor resend an already executed height?

dnkolegov: SubnetActor does not resend it. If SubnetActor sees the repeated checkpoint, it accumulates the relayer address and returns.

dnkolegov: This `else if` can be removed

src/gateway/GatewayRouterFacet.sol
dnkolegov: If reward is static and does not depends on messages  - we can

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Sorry, my comment was confusing. I just wanted to point out that here lies the problem that I was trying to explain yesterday: once the checkpoint is executed, the _active validator set_ changes and can only be used to validate the _next_ checkpoint, not another instance of the last one. 

We have the following options:
* We can only reward the first relayer
* We can maintain not one but two snapshots of the validator set: the one which validates the next, and the one which validates the last checkpoint
* Instead of validating the quorum signatures, we can compare the `checkpoint` to `s.committedCheckpoints[checkpoint.blockHeight]`, and ignore the signatures.

aakoshh: Unfortunately not maintaining previous snapshots makes it impossible to detect equivocations, but maybe that's a story for another day.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,52 +43,64 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+            // execute cross-messages
+            _applyMessages(checkpoint.subnetID, messages);
 
-        // TODO: distribute rewards to validators for their service executing a checkpoint
+            // reward relayers in the subnet for committing the previous checkpoint
+            // slither-disable-next-line unused-return
+            Address.functionCall(
+                msg.sender,
+                abi.encodeCall(ISubnetActor.rewardRelayers, (s.lastBottomUpExecutedCheckpointHeight, checkpointReward))
+            );
```
aakoshh, line 95: Does this really need to be here? `commitBottomUpCheckpoint` is called by the subnet actor itself, and I think it knows when it's a new checkpoint vs one that only needs to append a validator, so it should also be able to reward previous validators if that's how it works, right?

If it did the rewards on its own, the gateway wouldn't have to care at all how it's done. For example it could reward the first 3 relayers with diminishing returns (which was suggested by Guy). In that case it would be wrong for the gateway to trigger this at a later point.

src/subnet/SubnetActorManagerFacet.sol
aakoshh: And yes, it is trivial for lazy relayers to re-send a checkpoint they see on the parent subnet, without actually running a relayer that follows the child subnet. The only defense against that is to reward the first submitter a bit more, so there is some incentive to try to beat the crowd. 

Still, with the fixed pot distributed equally, at least there is no incentive to send under multiple accounts.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Sounds good.

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Fixed

src/subnet/SubnetActorManagerFacet.sol
adlrocha: Agree with @aakoshh. Let's have equal distribution for now to avoid sybils, and let's worry about the rest in the future. 

I see that this check has been fixed. Thanks!

src/gateway/GatewayRouterFacet.sol
```diff
@@ -41,35 +43,34 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
 
-        CrossMsg[] memory messages = s.bottomUpMessages[commit.blockHeight];
-
         uint256 totalValue = 0;
         uint256 crossMsgLength = messages.length;
         for (uint256 i = 0; i < crossMsgLength; ) {
+            // CODE-REVIEW
+            // before we used to do `totalValue += commit.fee + checkpoint.fee` here
+            // where and how are we going to use fees?
             totalValue += messages[i].message.value;
```
adlrocha, line 74: Here we are releasing `totalValue` from the circulating supply that is leaving the subnet when we `applyMessage` below. We should also release some funds to the subnet actor (this is why we had `fee` before) in order for the subnet to have enough funds to reward the relayers.

To summarize, I think we should include: 
- A `fee` field in `CrossMsgs` where users determine how much they want to reward the relayer for propagating their message.
- We should collect the amount and the `fee` from the balance of the addresses creating the message, when the checkpoint is submitted. 
- Fees are sent to the subnet actor so they can be distributed to relayers (if not, the subnet actor may not have enough funds to pay out relayers)

src/lib/LibStaking.sol
```diff
@@ -536,6 +550,20 @@ library LibStaking {
         emit CollateralClaimed(validator, amount);
     }
 
+    /// @notice method that allows a relayer to withdraw it's accumulated rewards using pull-based transfer
+    function claimRewardForRelayer(address relayer) external {
```
adlrocha, line 554: As mentioned in the comment above, we need to release some circulating supply from the gateway in order to have funds available for relayer payouts.

adlrocha (approved):
We are almost there. My last comment is related to how we provide the funds for rewards. The subnet actor needs to be provided with some of the circulating supply (in the for of fees) from the gateway in order to be able to reward relayers. Left a few comments around this. 

For me this would be the last thing pending in this PR, feel free to address it in the next PR if you see this one is becoming too noisy.

PR #230, opened 2023-10-06
Metadata with confirmation

cryptoAtwill:
Update the metadata through confirmation.


binding/src/gateway_diamond.rs
```diff
@@ -211,6 +211,7 @@ pub mod gateway_diamond {
     pub static GATEWAYDIAMOND_ABI: ::ethers::contract::Lazy<::ethers::core::abi::Abi> =
         ::ethers::contract::Lazy::new(__abi);
     #[rustfmt::skip]
+<<<<<<< HEAD
```
aakoshh, line 214: Merge conflict.

binding/src/subnet_actor_diamond.rs
```diff
@@ -227,11 +227,15 @@ pub mod subnet_actor_diamond {
     pub static SUBNETACTORDIAMOND_ABI: ::ethers::contract::Lazy<::ethers::core::abi::Abi> =
         ::ethers::contract::Lazy::new(__abi);
     #[rustfmt::skip]
+<<<<<<< HEAD
```
aakoshh, line 230: Merge conflict.

src/SubnetActorDiamond.sol
```diff
@@ -56,8 +57,13 @@ contract SubnetActorDiamond {
         s.majorityPercentage = params.majorityPercentage;
         s.currentSubnetHash = s.parentId.createSubnetId(address(this)).toHash();
 
-        //  We hardcode the current limit for active validators to 100 per Tendermint consensus
+        // We hardcode the current limit for active validators to 100 per Tendermint consensus
         s.validatorSet.activeLimit = 100;
```
aakoshh, line 61: Could we make this into a constructor argument?

src/lib/LibStaking.sol
```diff
@@ -5,62 +5,92 @@ import {IGateway} from "../interfaces/IGateway.sol";
...
+/// The util library for `StakingChangeLog`
+library LibStakingChangeLog {
+    event NewStakingChangeRequest(StakingOperation op, address validator, bytes payload, uint64 configurationNumber);
+
+    /// @notice Validator request to update its metadata
+    function metadataRequest(StakingChangeLog storage changes, address validator, bytes calldata metadata) internal {
+        uint64 configurationNumber = recordChange({
+            changes: changes,
+            validator: validator,
+            op: StakingOperation.Withdraw,
```
aakoshh, line 20: Is this `Withdraw` on purpose?

src/lib/LibStaking.sol
```diff
@@ -382,50 +423,76 @@ library LibStaking {
...
+    }
+
+    /// @notice Confirm the withdraw directly without going through the confirmation process
+    function withdrawWithConfirm(address validator, uint256 amount) internal {
+        SubnetActorStorage storage s = LibSubnetActorStorage.appStorage();
+
+        // record deposit that updates the total collateral
+        s.validatorSet.recordWithdraw(validator, amount);
+        // confirm deposit that updates the confirmed collateral
+        s.validatorSet.confirmWithdraw(validator, amount);
```
aakoshh, line 476:
```suggestion
        // record withdraw that updates the total collateral
        s.validatorSet.recordWithdraw(validator, amount);
        // confirm withdraw that updates the confirmed collateral
        s.validatorSet.confirmWithdraw(validator, amount);
```

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -78,8 +87,28 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
-        LibStaking.setValidatorMetadata(msg.sender, metadata);
-        LibStaking.deposit(msg.sender, msg.value);
+        if (!s.bootstrapped) {
+            // if the subnet has not been bootstrapped, join directly
+            // without delays, and collect collateral to register
+            // in the gateway
+
+            // confirm validators deposit immediately
+            LibStaking.depositWithConfirm(msg.sender, msg.value);
+            LibStaking.setMetadataWithConfirm(msg.sender, metadata);
```
aakoshh, line 97: This is where we discussed that we'll have to do https://github.com/consensus-shipyard/fendermint/issues/305  and that it will be as simple as checking that the `address(metadata) == msg.sender`.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -68,6 +71,12 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         if (!LibStaking.hasStaked(msg.sender)) {
             revert NotStakedBefore();
         }
+
+        if (!s.bootstrapped) {
+            LibStaking.setMetadataWithConfirm(msg.sender, metadata);
+            return;
+        }
+
```
aakoshh, line 79: We discussed with @adlrocha that for now we should simplify our life and have a 1-to-1 relationship between the validator public key and its address.

I think it would be best to disable `setMetadata` for this kind of consensus. I see that currently we have `Deposit | Withdraw | SetMetadata` all working with the sender address. If we keep it this way to avoid more churn, and we can guarantee that the `SetMetadata` is going to be the [first](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/230/files#diff-2bab5e461ace2b433d6865a2baf6846c211f59981f1956eca9b3026418a23af2R109-R110) thing to arrive at the child subnet, and then no more `SetMetadata` will follow to try to change what is the public key of the address, then we can stick to this. 

The other option as I see it is to reify the `metadata` in `join` as a `PublicKey` and use that as identifier for all subsequent operations.

aakoshh (commented):
The way you changed the `!bootstrap` decision looks good to me :+1: 

I am sorry about the back and forth with the metadata. Seems like a simple thing to get the damn public key into the child, but it just keeps spinning out of control, doesn't it? 

You have been iterating on it for a while now, so I'm prepared to go with however you want to propagate it to the child, but just to reiterate, what we need is actually the simplest of all designs:
* All validators need an Ethereum address, otherwise they can't call `join`. 
* They need to pass the PublicKey to `join`, so that it can go down to CometBFT in the child subnet. 
* We simplify a lot of book keeping and configuration malarkey if we agree that the PublicKey has to correspond to the Ethereum address. 
* Validators are free to create a dedicated account for the key they want to use to author blocks. 
* We can use the PublicKey to refer to the validator in every staking change.
* We don't have a need for the metadata at the moment (nor do I think we can trust it).
* But we can use the metadata as a mechanism as long as it can only be set during `join` and it's verified to be the public key of the account, and it's the update to arrive - at least for Fendermint subnets. 
* We can use the address to identify the validator in staking changes, as long as the PublicKey is already available in the child subnet when we make validator transitions.

src/lib/LibStaking.sol
cryptoAtwill: it's a bug, I fixed it already.

binding/src/subnet_actor_diamond.rs
cryptoAtwill: resolved

binding/src/gateway_diamond.rs
cryptoAtwill: resolved

src/subnet/SubnetActorManagerFacet.sol
cryptoAtwill: it's added

cryptoAtwill:
@aakoshh The check of metadata is added, please take a look again.

src/SubnetActorDiamond.sol
```diff
@@ -56,8 +58,13 @@ contract SubnetActorDiamond {
         s.majorityPercentage = params.majorityPercentage;
         s.currentSubnetHash = s.parentId.createSubnetId(address(this)).toHash();
 
-        //  We hardcode the current limit for active validators to 100 per Tendermint consensus
-        s.validatorSet.activeLimit = 100;
+        // We hardcode the current limit for active validators to 100 per Tendermint consensus
```
adlrocha, line 61: outdated comment.
```suggestion
        // We hardcode the current limit for active validators to 100 per Tendermint consensus
```

src/SubnetActorDiamond.sol
```diff
@@ -56,8 +58,13 @@ contract SubnetActorDiamond {
...
+        s.validatorSet.activeLimit = params.activeValidatorsLimit;
+        // Start the next configuration number from 1, 0 is reserved for no change.
```
adlrocha, line 63:
```suggestion
        // Start the next configuration number from 1, 0 is reserved for no change and the genesis membership
```

src/lib/LibStaking.sol
```diff
@@ -10,40 +10,61 @@ import {WithdrawExceedingCollateral, NotValidator, CannotConfirmFutureChanges, N
...
+            payload: metadata,
+            configurationNumber: configurationNumber
+        });
+    }
 
     /// @notice Perform upsert operation to the withdraw changes, return total value to withdraw
     /// @notice of the validator.
     /// Each insert will increment the configuration number by 1, update will not.
     function withdrawRequest(StakingChangeLog storage changes, address validator, uint256 amount) internal {
+        bytes memory payload = abi.encode(amount);
```
adlrocha, line 36: Why do you need to `abi.encode` the amount?

test/GatewayDiamond.t.sol
```diff
@@ -1798,15 +1799,15 @@ contract GatewayDiamondDeploymentTest is StdInvariant, Test {
         }
     }
 
-    function _join(address validatorAddress) internal {
-        vm.prank(validatorAddress);
-        vm.deal(validatorAddress, DEFAULT_COLLATERAL_AMOUNT + 1);
+    // function _join(address validatorAddress) internal {
```
adlrocha, line 1802: Should we remove this test or uncomment it?

adlrocha (approved):
LGTM. Just a few outstanding comments. 

I see that this PR conflicts with a few of the things I wanted to address next on my PR #228, mainly: 
* Remove the `f4` for validators. 
* Have public key explicitly in the validator struct instead of as part of the metadata 

I would suggest merging this one so I can merge this one and continue in #228?

src/lib/LibStaking.sol
adlrocha: As discussed sync, we are encoding all changes inside the payload to be more flexible with future changes. If we use this approach we need to have the encoding well-documented for users to be able to interpret what is happening easily.

src/SubnetActorDiamond.sol
```diff
@@ -56,8 +58,12 @@ contract SubnetActorDiamond {
...
         s.currentSubnetHash = s.parentId.createSubnetId(address(this)).toHash();
 
-        //  We hardcode the current limit for active validators to 100 per Tendermint consensus
-        s.validatorSet.activeLimit = 100;
+        s.validatorSet.activeLimit = params.activeValidatorsLimit;
+        // Start the next configuration number from 1, 0 is reserved for no change and the genesis membership
+        s.changeSet.nextConfigurationNumber = LibStaking.INITIAL_CONFIGURATION_NUMBER;
+        // The startConfiguration number is also 1 to match with nextConfigurationNumber, indicating we have
+        // empty validator change logs
+        s.changeSet.startConfigurationNumber = LibStaking.INITIAL_CONFIGURATION_NUMBER;
```
aakoshh, line 66: So `startConfigurationNumber` is the "first unapplied configuration"?

aakoshh (approved):
Thanks, looking good :+1:

test/GatewayDiamond.t.sol
adlrocha: As discussed sync, we will keep them commented until we come back to writing a better test coverage for the gateway.

src/SubnetActorDiamond.sol
cryptoAtwill: yeah...

PR #234, opened 2023-10-09
IPC-221: Bootstrap followup, clean outdated checkpoint, membership code

adlrocha:
Follow up #228

## Changes
This PR: 
- Removes outdated code from checkpoints, membership updates, and other logic that has been re-defined. 
- Removes the use of `FvmAddress` for validators. From now on, validators will only be allowed to use ETH (and their corresponding `f4`) addresses.
- Updates `Validator` types to use the same types in both, the subnet and gateway contracts. 
- Updates the gateway constructor to pass directly the genesis `Membership` so that also `metadata` is stored. 

## Follow up
In the next PR I will add the `applyChanges` method that Fendermint can use to pass a configuration number and trigger the update in membership in the gateway contract. I will also include additional unit tests and uncomment the end-to-end ones that can be fixed once that code is ready.


src/lib/LibGateway.sol
```diff
@@ -67,68 +66,74 @@ library LibGateway {
...
+            if (membership.configurationNumber < s.lastMembership.configurationNumber) {
+                revert OldConfigurationNumber();
+            }
 
-        s.lastMembership.configurationNumber = n;
-        // clear old validators
-        delete s.lastMembership.validators;
+            // Check if the memmbersip is equal and return if it is the case
+            if (membershipEqual(membership, s.currentMembership)) {
+                return;
```
aakoshh, line 87: So this is the case if there is a fresh configuration number but there is no real change? Why is it skipped, why not apply it whenever the number increases?

src/lib/LibGateway.sol
```diff
@@ -67,68 +66,74 @@ library LibGateway {
...
-            if (!FvmAddressHelper.equal(validators[i], zero)) {
-                uint256 validatorWeight = weights[i];
+        s.lastMembership = s.currentMembership;
 
-                if (validatorWeight == 0) {
-                    revert ValidatorWeightIsZero();
+        uint256 inputLength = membership.validators.length;
+        uint256 storLength = s.currentMembership.validators.length;
+        // memory arrays can't be copied directly from memory into storage,
+        // we need to explicitly increase the size of the array in storage.
```
aakoshh, line 96: What about decreasing it if the new array has fewer elements?

aakoshh: Never mind, found it further down.

src/lib/LibGateway.sol
adlrocha: This is not the `applyChange` function that you should call passing the config number. I have follow-up PR for that. This is just a convenient function to update the membership information and track the previous membership. 

The next PR will have a `applyFinality(configNumber)`  function that will confirm changes according to all of the ones included for the finality, update the membership in the contract state and return it to Fendermint (this is what you need right?)

src/lib/LibGateway.sol
```diff
@@ -67,68 +66,74 @@ library LibGateway {
...
+        for (uint256 i = 0; i < inputLength; ) {
+            s.currentMembership.validators[i] = membership.validators[i];
+            if (inputLength >= storLength) {
+                s.currentMembership.validators.push(membership.validators[i]);
+            }
```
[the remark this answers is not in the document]
adlrocha: Yes, sorry about that. I was doing some tests to make the Solidity compiler happy and forgot to revert :facepalm:

src/lib/LibGateway.sol
adlrocha: This is handled here, this is what you mean, right? https://github.com/consensus-shipyard/ipc-solidity-actors/pull/234/files#diff-565acf6db4298fb1595e6bca86a27eccfab40e5a1bd48d4a278282c11567a85bR108

aakoshh (approved):
Thanks, this should let me pass in genesis validators.

src/GatewayDiamond.sol
```diff
@@ -21,7 +21,6 @@ contract GatewayDiamond {
     struct ConstructorParams {
         SubnetID networkName;
         uint64 bottomUpCheckPeriod;
-        uint64 topDownCheckPeriod;
```
cryptoAtwill, line 24: ` using SubnetIDHelper for SubnetID;` at line 19 can also be deleted

adlrocha: fixed!

PR #235, opened 2023-10-10
FM-303: apply finality changes

adlrocha:
Closes: https://github.com/consensus-shipyard/fendermint/issues/303

## Changes
- [x] Method to apply changes committed in proofs-of-finality
- [x] Add `power_scale` to constructor in subnet actor so it is accessible for the child subnet in genesis in bootstrap.
- [x] Implement unit test for membership updates triggered after finality changes.


src/gateway/GatewayRouterFacet.sol
```diff
@@ -102,6 +103,38 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         s.validatorsTracker.batchStoreChange(changeRequests);
     }
 
+    /// @notice Apply all changes committed through the committment of parent finality
+    function applyFinalityChanges() external systemActorOnly returns (uint64) {
+        // get the latest configuration number for the change set
+        uint64 configurationNumber = s.validatorsTracker.changes.nextConfigurationNumber - 1;
```
dnkolegov, line 109: Small optimization. This read from the storage is not necessary in the next if statement is false. could be placed on line 117.

dnkolegov (approved):
Minor comments

src/SubnetActorDiamond.sol
```diff
@@ -44,6 +45,9 @@ contract SubnetActorDiamond {
         if (params.majorityPercentage < 51 || params.majorityPercentage > 100) {
             revert InvalidMajorityPercentage();
         }
+        if (params.powerScale > 12) {
```
aakoshh, line 48:
```suggestion
        if (params.powerScale > 18) {
```
https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/econ/mod.rs#L29

src/SubnetActorDiamond.sol
```diff
@@ -27,6 +27,7 @@ contract SubnetActorDiamond {
         uint64 bottomUpCheckPeriod;
         uint8 majorityPercentage;
         uint16 activeValidatorsLimit;
+        uint8 powerScale;
```
aakoshh, line 30:
```suggestion
        int8 powerScale;
```

src/gateway/GatewayRouterFacet.sol
aakoshh: If it has been accessed once, does it access it again? 

I'm asking because the `if` condition uses `s.validatorsTracker.changes.nextConfigurationNumber` too. Would it make sense to read it into a variable and only access the storage once?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -102,6 +103,38 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
+            address addr = validators[i];
+            ValidatorInfo storage info = s.validatorsTracker.validators.validators[addr];
+            vs[i] = Validator({weight: info.confirmedCollateral, addr: addr, metadata: info.metadata});
+            unchecked {
+                ++i;
+            }
+        }
+
+        // update membership with the applied changes
+        LibGateway.updateMembership(Membership({configurationNumber: configurationNumber, validators: vs}));
```
aakoshh, line 134: Do we still need the `Membership`? Isn't it possible to use just `listActiveValidators`? Although it is nice for snapshots, just want to understand what's the idea. Is the `validatorsTracker` an implementation detail and the membership is the public interface?

src/lib/LibSubnetActorStorage.sol
```diff
@@ -45,6 +45,9 @@ struct SubnetActorStorage {
     StakingChangeLog changeSet;
     /// @notice The staking release queue that only allow transfer of collateral after certain locking period.
     StakingReleaseQueue releaseQueue;
+    /// @notice Power scale determining the accuracy of the power scale (in number of decimals from whole FIL)
+    /// (e.g. miliFil = 3; microFIL = 6, attoFil = 12, etc.)
+    uint8 powerScale;
```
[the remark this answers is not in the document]
aakoshh: It can be equal to 0 for 1FIL=1power, but it can also be negative for 10FIL=1power.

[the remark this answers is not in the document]
aakoshh: It doesn't support 5FIL=1power :shrug:

aakoshh (commented):
Looks good apart from the Fendermint app allowing negative scale

src/SubnetActorDiamond.sol
adlrocha: Totally right, I don't know why I keep using `12` instead of `18` :facepalm:

src/gateway/GatewayRouterFacet.sol
adlrocha: Fixed so we have a single access. 
```
<rant> Solidity is a horrible programming language, reasoning about these things is impossible </rant> 
```
:laughing:

src/SubnetActorDiamond.sol
adlrocha: Why `int8`? `powerScale` is always positive, right?

src/gateway/GatewayRouterFacet.sol
adlrocha: Correct. While they are in sync and we could look at `listActiveValidators` in `validatorsTracker`, `currentMembership` and `lastMembership` are the public interfaces for current subnet membership. The reason for this is that `validatorsTracker` don't keep the previous membership, and IIRC, it was useful to keep this available in state.

src/SubnetActorDiamond.sol
adlrocha: Solved in https://github.com/consensus-shipyard/fendermint/pull/307#discussion_r1354617791

src/gateway/GatewayRouterFacet.sol
aakoshh: Sounds good!

aakoshh (approved):
Cool, now just waiting for the bindings to appear

PR #236, opened 2023-10-11
Add fee-based rewards

dnkolegov:


src/gateway/GatewayManagerFacet.sol
```diff
@@ -132,7 +132,8 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
             subnet: subnetId,
             signer: msg.sender,
             to: to,
-            value: msg.value - s.crossMsgFee
+            value: msg.value - s.crossMsgFee,
+            fee: s.crossMsgFee
```
adlrocha, line 136: Let's make this fee `0`. We decided that the fee for fund is going to be zero.

src/gateway/GatewayManagerFacet.sol
```diff
@@ -132,7 +132,8 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
             subnet: subnetId,
             signer: msg.sender,
             to: to,
-            value: msg.value - s.crossMsgFee
+            value: msg.value - s.crossMsgFee,
```
adlrocha, line 135: You can also remove `- s.crossMsgFee`

src/gateway/GatewayManagerFacet.sol
```diff
@@ -145,7 +146,8 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
             subnet: s.networkName,
             signer: msg.sender,
             to: to,
-            value: msg.value - s.crossMsgFee
+            value: msg.value - s.crossMsgFee,
```
adlrocha, line 149: I don't know like the way we do this. I was meaning to change it for a while. Instead of picking up the fee automatically  from the value given as an input (which could lead to a horrible UX), should we include an`fee` argument to the method so the user can determine the fee to pay? We can set a check here that verifies if the fee is larger than the minimum cross-msg fee for the subnet. Then the user or application (in our case the `ipc-cli`) will have to send `value + fee` in the `msg.value` so it can be deducted and the `fee` and amount that arrives to the destination is predictable.

Would you mind fixing it in this PR? Once this is merge I'll update the `ipc-cli` accordingly to include this argument. We can sync if this is not clear.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -66,28 +66,37 @@ contract GatewayRouterFacet is GatewayActorModifiers {
...
-            // where and how are we going to use fees?
             totalValue += messages[i].message.value;
+            totalFee += messages[i].message.fee;
             unchecked {
                 ++i;
             }
         }
 
-        if (subnet.circSupply < totalValue) {
+        uint256 totalCost = totalFee + totalValue;
```
adlrocha, line 79: No strong opinion, but `cost` sounds weird to me.
```suggestion
        uint256 totalAmount = totalFee + totalValue;
```

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -177,22 +174,26 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     }
 
     /// @notice reward the relayers for processing checkpoint at height `height`.
-    function rewardRelayers(uint64 height) internal {
-        if (s.relayerReward == 0) {
+    /// @dev The reward includes the fixed relayer reward and accumulated cross-message fees received from the gateway.
+    /// @param height height of the checkpoint the relayers are rewarded for
+    /// @param amount The sum of cross-message fees in the checkpint
+    function distributeRewardToRelayers(uint64 height, uint256 amount) external onlyGateway {
+        uint256 reward = s.relayerReward + amount;
```
adlrocha, line 181: I think we should remove the fixed relayer reward for now, and just distribute the fees that were committed in the checkpoint.

src/gateway/GatewayManagerFacet.sol
adlrocha: While you look into this, we need to be careful to burn the fees to in the child subnet when we do `commitBottomUp` (same with locking the fees in `topDown`. I don't think it changes anything if we follow the approach from above, but just in case I wanted to flag this to you: https://github.com/consensus-shipyard/ipc-solidity-actors/blob/a100008fc180538846ec1177f795a01aa0112a88/src/gateway/GatewayMessengerFacet.sol#L98

adlrocha (commented):
Amazing job, @dnkolegov. I few comments, but I don't think this one will require so much back and forth, it is really clear :pray:

src/gateway/GatewayManagerFacet.sol
dnkolegov: Fixed

src/gateway/GatewayManagerFacet.sol
dnkolegov: Fixed

src/gateway/GatewayRouterFacet.sol
dnkolegov: Fixed

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Remove or use set s.relayerReward to 0 ?

adlrocha: I think we should remove relayerReward for now (unless you strongly advocate otherwise). If we just reward the amount collected from the checkpoint, we know there will be enough balance in the subnet actor, bur `relayerReward` is an amount on top of what has been collected already, which would mess up the balance of the subnet actor, in many cases not being able to reward relayers at all. 

TL;DR: My take is to remove it all the way.

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -25,7 +25,7 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
      * @param crossMsg - a cross-message to send
      */
     function sendCrossMessage(CrossMsg calldata crossMsg) external payable hasFee {
-        if (crossMsg.message.value != msg.value - s.crossMsgFee) {
+        if (crossMsg.message.value != msg.value - s.minCrossMsgFee) {
```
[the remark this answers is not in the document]
adlrocha: You could maybe use this modifier? https://github.com/consensus-shipyard/ipc-solidity-actors/pull/236/files#diff-942c7c6d26ad2d36e81a825d97bb63ec692a92e68c6711ff2a603c2a54729275R94

src/gateway/GatewayManagerFacet.sol
```diff
@@ -132,20 +132,28 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
...
         LibGateway.commitTopDownMsg(crossMsg);
     }
 
     /// @notice release method locks funds in the current subnet and sends a cross message up the hierarchy to the parent gateway to release the funds
-    function release(FvmAddress calldata to) external payable hasFee {
+    function release(FvmAddress calldata to, uint256 fee) external payable hasFee {
+        if (fee < s.minCrossMsgFee) {
+            revert NotEnoughFee();
+        }
+        if (msg.value < fee) {
```
adlrocha, line 148: For this one you may be able to use this modifier: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/236/files#diff-942c7c6d26ad2d36e81a825d97bb63ec692a92e68c6711ff2a603c2a54729275R94

src/structs/Checkpoint.sol
```diff
@@ -60,4 +60,5 @@ struct StorableMsg {
     uint64 nonce;
     bytes4 method;
     bytes params;
+    uint256 fee;
```
adlrocha, line 63: Why do we need the `fee` here? `fees` are already included in the messages, and the messages are implicitly linked to the checkpoint through their hash, right? So having the `fee` here may be redundant, and even confusing.
```suggestion
```

adlrocha (changes requested):
A few changes more, and in the next one we should be good to go. Thanks!

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Fixed

src/structs/Checkpoint.sol
dnkolegov: That is the fee in the message.

src/gateway/GatewayManagerFacet.sol
dnkolegov: fixed

src/structs/Checkpoint.sol
adlrocha: Oh sorry, I thought this was inside the `Checkpoint` struct. This is GH messing with me. Thanks!

adlrocha (approved):
LGTM. Thank you!!

PR #238, opened 2023-10-12
Confirm membership and fix misprints

dnkolegov:
Confirming the next membership - https://github.com/consensus-shipyard/ipc-solidity-actors/blob/6322ae6a236859cedcc32d6ff7dabf4d709695e8/src/subnet/SubnetActorManagerFacet.sol#L73


adlrocha (approved):
Thank you @dnkolegov

PR #239, opened 2023-10-12
Support bootstrap nodes

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/fendermint/issues/312. It also makes several minor fixes related to getters and errors


src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -146,11 +145,17 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
     /// @notice method that allows a validator to leave the subnet
     function leave() external notKilled {
+        // remove bootstrap nodes added by this validator
+
         uint256 amount = LibStaking.totalValidatorCollateral(msg.sender);
         if (amount == 0) {
             revert NotValidator();
         }
 
+        // slither-disable-next-line unused-return
```
adlrocha, line 155: Here we are immediately removing the bootstrap instead of waiting for the leave to be effective for the validator, but I feel this is fine.

// cc @aakoshh @cryptoAtwill in case you have stronger opinions.

adlrocha (approved):
LGTM. Thanks. I'll add a command to the `ipc-cli` to add the validator. Once you have the script to deploy the bootstrap please let me know to see if I can also integrate it into the cli.

PR #240, opened 2023-10-13
Chain reorg detection

cryptoAtwill:
To help detect chain reorg, returning block hashes as required.

### DIFF
--- a/src/gateway/GatewayGetterFacet.sol
+++ b/src/gateway/GatewayGetterFacet.sol
@@ -82,14 +82,13 @@ contract GatewayGetterFacet {
 
     /// @notice get the list of top down messages from block number, we may also consider introducing pagination.
     /// @param subnetId - The subnet id to fetch messages from
-    /// @param fromBlock - The starting block to get top down messages, inclusive.
-    /// @param toBlock - The ending block to get top down messages, inclusive.
+    /// @param blockNumber - The block to get top down messages
     function getTopDownMsgs(
         SubnetID calldata subnetId,
-        uint256 fromBlock,
-        uint256 toBlock
-    ) external view returns (CrossMsg[] memory) {
-        return LibGateway.getTopDownMsgs({subnetId: subnetId, fromBlock: fromBlock, toBlock: toBlock});
+        uint256 blockNumber
+    ) external view returns (CrossMsg[] memory msgs, bytes32 blockHash) {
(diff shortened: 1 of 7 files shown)


src/gateway/GatewayGetterFacet.sol
```diff
@@ -82,14 +82,13 @@ contract GatewayGetterFacet {
...
-    /// @param fromBlock - The starting block to get top down messages, inclusive.
-    /// @param toBlock - The ending block to get top down messages, inclusive.
+    /// @param blockNumber - The block to get top down messages
     function getTopDownMsgs(
         SubnetID calldata subnetId,
-        uint256 fromBlock,
-        uint256 toBlock
-    ) external view returns (CrossMsg[] memory) {
-        return LibGateway.getTopDownMsgs({subnetId: subnetId, fromBlock: fromBlock, toBlock: toBlock});
+        uint256 blockNumber
```
adlrocha, line 88: This seems like a cosmetic change, can you explain a bit more the reasoning behind this for reorg detection?

adlrocha (approved):
LGTM. But do you mind explaining a bit the rationale behind the change before merging.

src/gateway/GatewayGetterFacet.sol
```diff
@@ -82,14 +82,13 @@ contract GatewayGetterFacet {
...
+    ) external view returns (CrossMsg[] memory msgs, bytes32 blockHash) {
+        msgs = LibGateway.getTopDownMsgs({subnetId: subnetId, fromBlock: blockNumber, toBlock: blockNumber});
+        blockHash = blockhash(blockNumber);
```
aakoshh, line 91: Is this a solidity built-in? Is it available all the way back to genesis? 

I did not expect that we need changes in Solidity for this, but I suspected the "by height" queryies in Ethereum don't return the identity of the block, which is why I mentioned the other option of querying "by hash", in which case this would not be needed. I'll check the agent PR to understand.

aakoshh: https://docs.soliditylang.org/en/latest/units-and-global-variables.html

> blockhash(uint blockNumber) returns (bytes32): hash of the given block when blocknumber is one of the 256 most recent blocks; otherwise returns zero

adlrocha: I also had a hard time understanding this change, @aakoshh comment reinforces me that we should revert this PR, and get topDownMessages for a specific height off-chain. This should be possible, right?

Issue #233, opened 2023-10-09
Separate tests and updating bindings tasks in the github workflows


dnkolegov:
CI/CD tests run a very long time due to implementing rust bindings in the same task/workflow.

dnkolegov:
Fixed in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/245

PR #245, opened 2023-10-14
Fix slither findings

dnkolegov:
It fixes several Slither findings and https://github.com/consensus-shipyard/ipc-solidity-actors/issues/233


adlrocha (approved):
I am fine by this, but can you explain the rationale behind the changek please? Why did you move it into its own workflow?

PR #250, opened 2023-10-17
Integration testing

aakoshh:
Includes fixes for problems discovered in https://github.com/consensus-shipyard/fendermint/pull/299 : 
* Take 65 long public keys
* Fix the hashing of public keys


src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -254,8 +259,12 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
 
-    function publicKeyToAddress(bytes memory publicKey) internal pure returns (address) {
-        bytes32 hashed = keccak256(publicKey);
-        return address(uint160(bytes20(hashed)));
+    /**
+     * @notice Hash a 65 byte public key and return the corresponding address.
+     */
+    function publicKeyToAddress(bytes calldata publicKey) internal pure returns (address) {
+        assert(publicKey.length == 65);
+        bytes32 hashed = keccak256(publicKey[1:]);
```
adlrocha, line 267: Why do we disregard the first byte?

adlrocha (approved):
@aakoshh, I fixed the tests, feel free to merge if they LGTY. Thanks!

src/subnet/SubnetActorManagerFacet.sol
aakoshh: The public key consists of two 32byte numbers, and it can be serialized in multiple ways:
* in full format, which is the concatentation of the two, or
* in compressed format, which is just 32 bytes

The 65 byte consists of a prefix byte that indicates whether we are dealing with the full or compressed versions, but once you say the key has to be 64 bytes long, it implies full format, so Ethereum decided to leave out the prefix. 

Here's how the [parsing](https://github.com/paritytech/libsecp256k1/blob/master/src/lib.rs#L122) works, showing all the options.

PR #252, opened 2023-10-19
Return previous finality

cryptoAtwill:
When committing parent finality, return the previous finality. This is for fendermint to validate the finalities are consistent.

### DIFF
--- a/src/gateway/GatewayRouterFacet.sol
+++ b/src/gateway/GatewayRouterFacet.sol
@@ -99,6 +99,7 @@ contract GatewayRouterFacet is GatewayActorModifiers {
     }
 
     /// @notice commit the ipc parent finality into storage and returns the previous committed finality
+    /// This is useful to understand if the finalities are consistent or if there have been reorgs.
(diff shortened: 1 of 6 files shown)


src/gateway/GatewayRouterFacet.sol
```diff
@@ -98,10 +98,13 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         });
     }
 
-    /// @notice commit the ipc parent finality into storage
+    /// @notice commit the ipc parent finality into storage and returns the previous committed finality
```
adlrocha, line 101:
```suggestion
    /// @notice commit the ipc parent finality into storage and returns the previous committed finality
    /// This is useful to understand if the finalities are consistent or if there have been reorgs.
```

adlrocha (approved):
Nice one. A slight suggestion to add come context of why do we need this, but LGTM

cryptoAtwill:
When the finality is committed, it returns the previous committed finality, so that we know from where to fetch the new messages.

PR #255, opened 2023-10-20
Integration testing fixes

aakoshh:
Further fixes for problems unearthed by https://github.com/consensus-shipyard/fendermint/pull/319

* Removed the `lastExecutedBottomUpCheckpoint` from `LibGatewayActorStorage`. Such a variable would need to be tracked per subnet.
* Ignore configuration number 0 in checkpoints
* Do not forget validator in withdrawal if it still has unconfirmed operations
* Release stake during checkpointing, not when the validator comes to claim]
* First commit the checkpoint to the gateway, then confirm the changes, because they might deactive the checkpoint, which would cause the commit to revert
* Do not call sink&swim when the last item of a priority queue is deleted
* Calculate the active collateral and use it for quorum checks instead of the total confirmed value


adlrocha (commented):
You caught a few bugs in this one, thank you @aakoshh. 

@dnkolegov, do you mind having a look to this one, and maybe fixing my comments and the tests so @aakoshh can focus on the Fendermint side of things? Thanks!

src/gateway/GatewayRouterFacet.sol
```diff
@@ -93,7 +87,7 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         // slither-disable-next-line unused-return
         Address.functionCallWithValue({
             target: msg.sender,
-            data: abi.encodeCall(ISubnetActor.distributeRewardToRelayers, (previousCheckpointHeight, totalFee)),
+            data: abi.encodeCall(ISubnetActor.distributeRewardToRelayers, (checkpoint.blockHeight, totalFee)),
```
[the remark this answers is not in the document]
aakoshh: I saw you commented on the this below, but just for clarity:
* I wanted to remove this call but couldn't, because the `totalFee` was calculated here
* I changed the `distributeRewardToRelayers` itself to give rewards to `n-1` instead of the height included in the call
* but it pays out the reward calculated to `n` at the moment, so the relayers of `n-1` get their rewards from the fees arriving in `n`
* I didn't know this was the way fees were distributed. I thought we would give out a fixed fee for relayers, and the cross message fees would go to validators, and the per-message reimbursement for the relayer triggering the execution on Lotus came from the gas accounting done by the gateway
* but I didn't have a better idea, just wanted to make it work

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -201,21 +201,26 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
     /// @dev The reward includes the fixed relayer reward and accumulated cross-message fees received from the gateway.
     /// @param height height of the checkpoint the relayers are rewarded for
     /// @param reward The sum of cross-message fees in the checkpoint
     function distributeRewardToRelayers(uint64 height, uint256 reward) external onlyGateway {
         if (reward == 0) {
             return;
         }
-        address[] memory relayers = s.rewardedRelayers[height].values();
+        uint64 previousHeight = height - s.bottomUpCheckPeriod;
+        address[] memory relayers = s.rewardedRelayers[previousHeight].values();
```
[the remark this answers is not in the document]
aakoshh: Not every time, this only gets done the first time a checkpoint is submitted, see [here](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/f33f88dccfea7a7ee6435b1e0b4e50093990a8c6/src/subnet/SubnetActorManagerFacet.sol#L58-L87).

src/gateway/GatewayRouterFacet.sol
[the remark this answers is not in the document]
adlrocha: Then this is working as intended, thanks!

src/subnet/SubnetActorManagerFacet.sol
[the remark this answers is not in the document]
adlrocha: You are right, the changeset was not showing the whole code and for some reason I decided to look only into the subnet contract. Thanks!

src/gateway/GatewayRouterFacet.sol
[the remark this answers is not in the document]
aakoshh: Well, in theory you could save the rewards for later, look up the previous one from storage. 

The way it is will actually lead to permanent locking up of some rewards: 
0. Say our checkpoint interval is `C`
1. Start in genesis from height 0. 
2. The first checkpoint arrives at height `C`. 
3. Sum up the cross message fees in the messages, call it `R`
4. Look up the relayers of the previous checkpoint `C-C`, at height 0, which is an empty list
5. Skip distributing `R`, effectively locking them forever in the subnet actor or gateway

So the rewards for the first checkpoint go nowhere because there is no earlier checkpoint :shrug: 

Instead, the execution of `C` could store `R` at height `C`, and then the _next_ checkpoint could look it up and dish it out to the all the relayers of `C`.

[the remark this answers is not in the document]
aakoshh: What I wanted to say was: I'm not sure this is the way it was actually intended to work, or it's just an accident of the fact that I suggested rewarding not just the first but potentially all relayers of a checkpoint, which means holding on to rewards until we know who they are, even after the first one caused the checkpoint to be committed.

src/lib/LibGatewayActorStorage.sol
```diff
@@ -72,8 +72,6 @@ struct GatewayActorStorage {
     uint64 bottomUpCheckPeriod;
     /// Tracking validator changes from parent in child subnet
     ParentValidatorsTracker validatorsTracker;
-    /// @notice The height of the last received bottom-up checkpoint.
-    uint64 lastBottomUpExecutedCheckpointHeight;
```
dnkolegov, line 76: > Such a variable would need to be tracked per subnet.

Where is the code implementing that tracking per subnet?

src/lib/LibStaking.sol
```diff
@@ -645,6 +648,8 @@ library LibValidatorTracking {
     function confirmChange(ParentValidatorsTracker storage self, uint64 configurationNumber) internal {
         if (configurationNumber >= self.changes.nextConfigurationNumber) {
             revert CannotConfirmFutureChanges();
+        } else if (configurationNumber < self.changes.startConfigurationNumber) {
+            return;
```
dnkolegov, line 652: Why do you `return` but not `revert` here?

dnkolegov:
> Ignore configuration number 0 in checkpoints

@aakoshh Where is that implemented?

src/lib/LibGatewayActorStorage.sol
aakoshh: There isn't one. Every time a checkpoint is executed I assume it follows the previously executed checkpoint by exactly one period.

src/lib/LibStaking.sol
aakoshh: That's the answer to https://github.com/consensus-shipyard/ipc-solidity-actors/pull/255#issuecomment-1774780657 

We can revert, I just didn't want to introduce another error. The point was to ignore 0, which is possibly just a confusing way of signaling "no change". We can revert on any non-zero value that isn't the expected one.

src/lib/LibStaking.sol
```diff
@@ -588,6 +589,7 @@ library LibStaking {
                 if (change.op == StakingOperation.Withdraw) {
                     s.validatorSet.confirmWithdraw(validator, amount);
                     s.releaseQueue.addNewRelease(validator, amount);
+                    IGateway(s.ipcGatewayAddr).releaseStake(amount);
                 } else {
                     s.validatorSet.confirmDeposit(validator, amount);
                     IGateway(s.ipcGatewayAddr).addStake{value: amount}();
```
aakoshh, line 595: I see that in `releaseStake` the gateway sends tokens to the subnet actor. Where is the opposite transfer?

adlrocha: You mean the `addStake`? Or the `register` with the initial allocation of collateral to the gateway from the subnet actor?

aakoshh: I mean that `releaseStake` sends tokens from the gateway back to the subnet actor, but on the `else` branch I don't see the opposite happening, of the subnet actor sending value to the gateway that is later returned.

adlrocha: Can you point me to the `else` that you are refering to? This is where we add the funds the gateway that are then picked up in `releaseStake`: https://github.com/consensus-shipyard/ipc-solidity-actors/blob/93d8e35a7fc861987bcab1f928ec5de5eeb29c90/src/lib/LibStaking.sol#L593

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -201,21 +201,26 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
         uint256 relayersLength = relayers.length;
         if (relayersLength == 0) {
             return;
         }
         if (reward < relayersLength) {
-            revert NotEnoughBalanceForRewards();
+            // Reverting here would mean a single message with 1 atto reward
+            // relayed by 2 validators would mean the checkpoint cannot be
+            // submitted.
+            // revert NotEnoughBalanceForRewards();
```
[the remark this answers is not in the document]
adlrocha: Let's remove the revert line to avoid confusions?

adlrocha (approved):
Overall LGTM, thank you @aakoshh.

@cryptoAtwill, once this is merged this is the version you should use for testing.
@dnkolegov, can you do a final pass before merging? Thanks!

src/lib/LibStaking.sol
aakoshh: Ahh, thanks, I missed the `{value: amount}` part :see_no_evil:

PR #257, opened 2023-10-23
Add unstake method

dnkolegov:
This PR fixes https://github.com/consensus-shipyard/ipc-solidity-actors/issues/256


src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -130,7 +130,7 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         }
     }
 
-    /// @notice method that allows a validator to increase their stake
+    /// @notice method that allows a validator to increase its stake
```
aakoshh, line 133: FWIW I think _their_ was the correct word. It can be used to refer to a person without picking a gender, while _it_ is more for animals and objects.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -148,6 +148,15 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         LibStaking.deposit(msg.sender, msg.value);
     }
 
+    /// @notice method that allows a validator to unstake a part of its collateral from a subnet
+    function unstake(uint256 amount) external notKilled {
+        if (LibStaking.totalValidatorCollateral(msg.sender) <= amount) {
+            revert NotEnoughCollateral();
+        }
```
aakoshh, line 155: To be consistent with `leave` this could return `NotValidator` like [here](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/4e2734285379bb9a92e696e1bcbc8bf1ac30b316/src/subnet/SubnetActorManagerFacet.sol#L165-L167) if the collateral is 0, and `NotEnoughCollateral` otherwise. Or it could unstake everything up to the current amount.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -148,6 +148,15 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+
+        LibStaking.withdraw(msg.sender, amount);
```
aakoshh, line 157: To be consistent with `leave`, this could first check if bootstrapping has been completed already, and if not then return immediately, like [here](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/4e2734285379bb9a92e696e1bcbc8bf1ac30b316/src/subnet/SubnetActorManagerFacet.sol#L173-L176).

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -148,6 +148,15 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         LibStaking.deposit(msg.sender, msg.value);
     }
 
+    /// @notice method that allows a validator to unstake a part of its collateral from a subnet
+    function unstake(uint256 amount) external notKilled {
```
aakoshh, line 152: One reason I didn't add this method myself was that I didn't want to go into finessing about what should happen when somebody wants to unstake all their stake. If they call `leave` it removes them from the `bootstrapOwners` as well. I think calling `leave()` and calling `unstake` with all remaining collateral should have the same semantics.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -148,6 +148,15 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        if (LibStaking.totalValidatorCollateral(msg.sender) <= amount) {
```
aakoshh, line 153: Ah, do you disallow unstaking all so they have to call leave?

aakoshh (commented):
Thank you for implementing this. FWIW in the tests I implemented more permissive rules where trying to unstake more than what you had resulted in getting back all outstanding amount. 

I don't have a strong opinion either way. @adlrocha what do you think?

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: ok

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Fixed

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Fixed

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Yes, that was the idea.

aakoshh (approved):
Thanks, this looks good to me, but let's give @adlrocha  a chance to see if he agrees with the no-leave approach.

adlrocha:
> I don't have a strong opinion either way. @adlrocha what do you think?

I don't have a strong opinion. I rather be conservative here so if a user does `unstake(2000)` when their collateral was just `1000` and they were looking to unstake `200` it should fail preventing "fat finger blunders". I am used to make several of these, so that is why I am more inclined to this  :)

That being said, if no one has a strong opinion let's get this one merged and improve from user feedback. Thank you both!

PR #259, opened 2023-10-24
Unit tests - Part 1

dnkolegov:
This is the first part of the tests.


adlrocha (approved):
I just skimmed through the tests, let me know if there is a specific one you want me to have a deeper look at. 

Before we keep writing tests, a few comments being also discussed in [this thread](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1698323427100699?thread_ts=1698250605.780749&cid=C04JR5R1UL8).
* I think we should remove every mock contract and integrate instances of the real versions of the subnet and gateway actor as we had originally. If we see there is a lot of redundant code, we can have a single set of tests that test the behavior of both contracts, but I would avoid mocked tests as they may end up being hard to maintain and they are not real.
* We should maybe think through the best way to write these tests so that you don't have to come up yourself with every possible scenario having to do a lot of monkey work

dnkolegov:
> I just skimmed through the tests, let me know if there is a specific one you want me to have a deeper look at.
> 
> Before we keep writing tests, a few comments being also discussed in [this thread](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1698323427100699?thread_ts=1698250605.780749&cid=C04JR5R1UL8).
> 
> * I think we should remove every mock contract and integrate instances of the real versions of the subnet and gateway actor as we had originally. If we see there is a lot of redundant code, we can have a single set of tests that test the behavior of both contracts, but I would avoid mocked tests as they may end up being hard to maintain and they are not real.
> * We should maybe think through the best way to write these tests so that you don't have to come up yourself with every possible scenario having to do a lot of monkey work

Thank you @adlrocha.
I agree with you, but then I do not understand why we switched to mocks, and now we will switch to real contractors back.

PR #260, opened 2023-10-24
FIX: Cannot unstake 0

aakoshh:
Fixes an error discovered by https://github.com/consensus-shipyard/fendermint/pull/319
Follow up to https://github.com/consensus-shipyard/ipc-solidity-actors/pull/257

The reason we can't allow 0 in unstake is because if we do, the next checkpoint fails:

```
checkpoint submission should succeed: CallError { exit_code: ExitCode { value: 33 }, failure_info: Some(MessageBacktrace(Backtrace { frames: [Frame { source: 64, method: 6, code: ExitCode { value: 33 }, message: "contract reverted" }, Frame { source: 64, method: 3844450837, code: ExitCode { value: 33 }, message: "contract reverted" }, Frame { source: 120, method: 6, code: ExitCode { value: 33 }, message: "contract reverted" }, Frame { source: 120, method: 3844450837, code: ExitCode { value: 33 }, message: "contract reverted" }], cause: None })), error: fendermint_vm_actor_interface::ipc::subnet::SubnetActorErrors:GatewayManagerFacetErrors(CannotReleaseZero(CannotReleaseZero)) }

```

### DIFF
--- a/src/subnet/SubnetActorManagerFacet.sol
+++ b/src/subnet/SubnetActorManagerFacet.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity 0.8.19;
 
-import {CollateralIsZero, NotOwnerOfPublicKey, EmptyAddress, NotEnoughBalanceForRewards, NotEnoughCollateral, NotValidator, NotAllValidatorsHaveLeft, NotStakedBefore, InvalidSignatureErr, InvalidCheckpointEpoch, InvalidCheckpointMessagesHash, InvalidPublicKeyLength} from "../errors/IPCErrors.sol";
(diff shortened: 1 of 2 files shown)


src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -152,10 +152,16 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     /// @notice method that allows a validator to unstake a part of its collateral from a subnet
     /// @dev `leave` must be used to unstake the entire stake.
     function unstake(uint256 amount) external notKilled {
-        if (LibStaking.totalValidatorCollateral(msg.sender) == 0) {
+        if (amount == 0) {
+            revert CollateralIsZero();
```
adlrocha, line 156: Maybe use https://github.com/consensus-shipyard/ipc-solidity-actors/blob/cc109a57d62755cf849028a366258c38c04e49c8/src/errors/IPCErrors.sol#L7 instead?
```suggestion
            revert CannotReleaseZero();
```

adlrocha (approved):
Really good catch, the gateway prevents zero releases to prevent useless calls. Unfortunately, with the new staking module, this call has been hidden deep into the code :) Thanks!

src/subnet/SubnetActorManagerFacet.sol
aakoshh: Yeah I also thought whether it should be this, but then I thought it also makes sense to say "the collateral you want to unstake is zero!". Will change.

PR #261, opened 2023-10-25
Lib staking bindings

cryptoAtwill:
Expose `LibStaking` and `LibStakingChangeLog` bindings so that `ipc-provider` can query the `NewStakingRequest` event.

### DIFF
--- a/src/lib/LibStakingChangeLog.sol
+++ b/src/lib/LibStakingChangeLog.sol
@@ -1,0 +1,92 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.8.19;
+
+import {StakingChangeLog, StakingChange, StakingOperation} from "../structs/Subnet.sol";
+
+/// The util library for `StakingChangeLog`
+library LibStakingChangeLog {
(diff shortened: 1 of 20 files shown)


src/lib/LibStakingChangeLog.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity 0.8.19;
```
adlrocha, line 2: Out of a curiosity, why did you decide to move this into its own library? I am fine with this, it is cleaner, but would love to understand the reasoning behind it :)

adlrocha (approved):
Good catch, lgtm. Thanks!

src/lib/LibStakingChangeLog.sol
cryptoAtwill: It's because the `build.rs` requires it to be in a separate file.

cryptoAtwill: no choice if we dont change build.rs and it's quite simple to change.

PR #262, opened 2023-10-26
Mikers/diamond upgrades

snissn:
- used the DiamondCutFacet contract in order to make the diamondCut method accessible to the contract owner for future upgrades
- added enforceIsContractOwner to the diamond library LibDiamond which exists in other versions of the library
- added replace and remove functionality to LibDiamond and relevant error methods
- added diamondCut to the gateway deploy template
- added diamondCut to subnet actor deploy script
- added tests to gateway and subnet actor for adding functions, replacing functions, deleting functions and asserting all of that is available only to the owner


src/diamond/DiamondCutFacet.sol
```diff
@@ -0,0 +1,23 @@
...
+contract DiamondCutFacet is IDiamondCut {
+    /// @notice Add/replace/remove any number of functions and optionally execute
+    ///         a function with delegatecall
+    /// @param _diamondCut Contains the facet addresses and function selectors
+    /// @param _init The address of the contract or facet to execute _calldata
+    /// @param _calldata A function call, including function selector and arguments
+    ///                  _calldata is executed with delegatecall on _init
+    function diamondCut(FacetCut[] calldata _diamondCut, address _init, bytes calldata _calldata) external override {
+        LibDiamond.enforceIsContractOwner();
+        LibDiamond.diamondCut({_diamondCut: _diamondCut, _init: _init, _calldata: _calldata});
```
adlrocha, line 21: This `_calldata` could be used to trigger the upgrade of some other contract if needed, right?

adlrocha (commented):
Thank you @snissn, overall LGTM, but I would like @dnkolegov to have a final look before merging in case I missed something, as he was the one implementing the original diamond architecture. 

As next steps I would suggest: 
* Writing a script exposed through `Makefile` that updates the facets of the gateway contract to a new version. The script should check if there are changes in the facet, and deploy those that have changed to the new version (if this is not easy to do, we can also deploy all the facets from scratch even if it is more expensive).
* With the above, we should understand what would it take to trigger this from Rust for the subnet actor so we can expose a `subnet upgrade` function similar to `subnet create` from the `ipc-cli`: https://github.com/consensus-shipyard/ipc/blob/9186517723e843ba894ba0a6781d0d1ff4dabe25/ipc/provider/src/manager/evm/manager.rs#L207

As a bonus task, we should port the registry to a diamond architecture so when there is an update to the reference implementation of the subnet actor we update the factory to use the latest version (honestly, this is also a bonus because I don't know to what extent this is possible, I haven't had the time to look into it)

dnkolegov (approved):

src/diamond/DiamondCutFacet.sol
snissn: yes! definitely -- so diamondCut's first argument has all of the changes that need to be made to the inner workings of the diamond structure, and then `_init.delegatecall(_calldata);` is called which is where we can call an arbitrary contract with arbitrary data.

PR #264, opened 2023-10-27
Unit test - part 2

dnkolegov:
This PR adds more tests and fixes several bugs.


dnkolegov:
@aakoshh Could you verify this one https://github.com/consensus-shipyard/ipc-solidity-actors/pull/264/commits/c1c69286358b4377cdd70114577fc2917992f927#diff-565acf6db4298fb1595e6bca86a27eccfab40e5a1bd48d4a278282c11567a85bR112 ?

I found that configuration numbers were not updated in the contracts. Do you have similar tests in your framework?

src/lib/LibGateway.sol
```diff
@@ -109,6 +109,7 @@ library LibGateway {
                 ++i;
             }
         }
+        s.currentMembership.configurationNumber = membership.configurationNumber;
```
adlrocha, line 112: Was this a bug where the config number wasn't being updated? (I am asking to get some context on the changes, the fact that all the code changes are in the same PR makes it hard to discern what was a bug or something else). Thanks!

adlrocha (approved):

src/lib/LibGateway.sol
dnkolegov: Correct

PR #265, opened 2023-10-30
Check sender when adding a signature

dnkolegov:
It is possible that someone could have taken another validator’s signature and called `addSignature` function.
We should enforce `msg.sender == recoveredSignatory`.


aakoshh:
I don't think it's necessary to restrict it so that the key who signed the checkpoint is the same as the one who sent the message. Ostensibly validators could use a different account to broadcast their transactions [here](https://github.com/consensus-shipyard/fendermint/blob/05cd5e77721a44bd5d0665edd2052762937fc0be/fendermint/app/src/cmd/run.rs#L92-L97). We are using the same thinking that perhaps this is a way to prioritise them, but why would it be a problem is someone else relayed a validator's signature to us?

dnkolegov:
> but why would it be a problem is someone else relayed a validator's signature to us?

Because it is incorrect in our system:

- We do not have such a requirement
- Our design does not consider that feature explicitly
- Maybe the signatory did not want to send it, or maybe the signatory wanted to send it later
- When someone sends a signature it stays completely anonymous in our system

So I think we should not allow sending signatures from anonymous parties until we need it and design it

aakoshh:
I'd say once the signatory has signed the checkpoint, and that signature has been made known to another party, there's nothing they can do about it. 

You are saying that maybe they did not want to send it to the child ledger (yet). But the child ledger is only an accumulation area for signatures. Someone who has the signature can always add it to the list of signatures which is being relayed to the parent subnet, and if it checks out, then there is nothing at the moment that the signatory can do to rescind it. 

So while it's not an explicit requirement, there's also no harm coming from it, and I still don't see why it should be considered illegal to submit someone else's to the list.

dnkolegov:
It is closed according to the Akosh's comments.

PR #266, opened 2023-10-30
Remove check

cryptoAtwill:
Remove the check on `checkpointInfo.threshold` due to:
- 0 threshold is also valid threshold.
- Checkpoint info is created together with checkpoint, as long as checkpoint is created, checkpoint info will be created. Checking checkpoint should be enough.

### DIFF
--- a/src/gateway/GatewayRouterFacet.sol
+++ b/src/gateway/GatewayRouterFacet.sol
@@ -12,7 +12,7 @@ import {IPCMsgType} from "../enums/IPCMsgType.sol";
 import {Membership} from "../structs/Subnet.sol";
 import {NotEnoughSubnetCircSupply, InvalidCheckpointEpoch, InvalidSignature, NotAuthorized, SignatureReplay, InvalidRetentionHeight, FailedRemoveIncompleteCheckpoint} from "../errors/IPCErrors.sol";
 import {InvalidCheckpointSource, InvalidCrossMsgNonce, InvalidCrossMsgDstSubnet, CheckpointAlreadyExists, CheckpointInfoAlreadyExists, CheckpointAlreadyProcessed, FailedAddIncompleteCheckpoint, FailedAddSignatory} from "../errors/IPCErrors.sol";
-import {NotEnoughBalance, NotRegisteredSubnet, SubnetNotActive, SubnetNotFound, InvalidSubnet, CheckpointNotCreated, CheckpointMembershipNotCreated, ZeroMembershipWeight} from "../errors/IPCErrors.sol";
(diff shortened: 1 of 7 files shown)


src/gateway/GatewayRouterFacet.sol
```diff
@@ -231,9 +231,6 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         }
 
         CheckpointInfo storage checkpointInfo = s.bottomUpCheckpointInfo[height];
-        if (checkpointInfo.threshold == 0) {
```
dnkolegov, line 234: The goal of this check was to verify that a checkpoint had been created. Before that, we suggested that the threshold cannot be equal to 0; in that case, equality to 0 meant that the checkpoint did not exist.
If the threshold can be equal to 0, then you need to implement another mechanism that allows you to verify the existence of checkpoint and checkpoint information.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -307,9 +304,6 @@ contract GatewayRouterFacet is GatewayActorModifiers {
         if (s.bottomUpCheckpoints[checkpoint.blockHeight].blockHeight > 0) {
             revert CheckpointAlreadyExists();
         }
-        if (s.bottomUpCheckpointInfo[checkpoint.blockHeight].threshold > 0) {
```
dnkolegov, line 310: The same comment from above is applied here.
You cannot guarantee that separate data types are managed synchronously. We need a mechanism to validate that.

You can combine those types or add a new field signaling whether checkpoint/info exists or not.

dnkolegov (changes requested):

src/gateway/GatewayRouterFacet.sol
cryptoAtwill: so effectively, we can create some kind of lib or wrapper to manage these two together? Because what happens now is the existence of checkpoint implies the existence of checkpoint info.

dnkolegov: I do not think we need an additional library for that. 
And I do not see any problems with your last statement either.

It is enough to add a bool field in both structs and check its value instead of checking `threshold`

aakoshh: It's true though that at the moment both of them are accessed together every time: inserted, read and deleted in unison. A wrapper with 2 fields (checkpoint and info) could save an extra dictionary traversal.

adlrocha (approved):
LGTM, thanks. The code looks cleaner to me know. I would let @dnkolegov, have a final pass before merging as I know he had some concerns (that being said looks fine for me).

dnkolegov (approved):

PR #268, opened 2023-10-31
1 - upgrade facets

snissn:
use hardhat to upgrade facets attached to the gateway


snissn:
https://filecoinproject.slack.com/files/U042VH0A17Y/F063PUS8HB5/diamond_gateway_upgrade_2.mov

here's a demo of how the hardhat command `npx hardhat upgrade-gw-diamond works` - i simulate making a change to some of the code in a function in the GatewayRouterFacet to simulate a bugfix, run the hardhat command and deploy the function changes, then I revert the changes locally, run the command again and it updates it again. i also show that it does nothing when called without any changes

scripts/util.ts
```diff
@@ -1,25 +1,258 @@
...
+async function stopGanache(server) {
+    return new Promise((resolve, reject) => {
+        server.close((err) => {
+            if (err) reject(err)
+            else resolve()
+        })
+    })
+}
+
+export async function getRuntimeBytecode(bytecode) {
```
adlrocha, line 109: I see, so in order to compare the bytecode deployed to the one locally, you do a test deployment using ganache. Do you mind adding a description of the process of upgrading in `updade-sa-diamond` and `upgrade-gw-diamond`, or even better, add a description in the README? I feel this could really help someone coming in the future to edit these scripts. Thanks :pray:

scripts/upgrade-gw-diamond.ts
```diff
@@ -0,0 +1,62 @@
...
+    //return this object to update the caller on what facets where updated
+    const updatedFacets = {}
+
+    const onChainFacetBytecodes = await getOnChainBytecodeFromFacets(facets)
+
+    // Loop through each facet in the deployments
+    for (let facetIndex in deployments.Facets) {
+        const facet = deployments.Facets[facetIndex]
+        const facetBytecode = await getBytecodeFromFacet(facet)
+        if (!onChainFacetBytecodes.hasOwnProperty(facetBytecode)) {
```
adlrocha, line 32:
```suggestion
       // compare the bytecode of the facets deployed with the local version of the contract.
        if (!onChainFacetBytecodes.hasOwnProperty(facetBytecode)) {
```

scripts/util.ts
```diff
@@ -1,25 +1,258 @@
...
+    await stopGanache(ganacheServer)
+
+    return runtimeBytecode
+}
+
+export async function getBytecodeFromFacet(facet) {
+    const facetName = facet.name
+    const libs = facet.libs
+    const bytecodeNeedsLink = getBytecodeFromFacetTypeChainFilename(
+        `./typechain/factories/${facetName}__factory.ts`,
```
adlrocha, line 133: Who (or what) generates this file?

adlrocha (commented):
Amazing job! I left a few suggestions. Also, would you mind adding a description of how the upgrade process works in the README? It would be useful to share some context of what the scripts do for someone coming new into the code base. Thanks!

scripts/upgrade-gw-diamond.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const lazyImport = async (module: any) => {
+    return await import(module)
+}
+
+
+// Function to upgrade the Subnet Actor Diamond
+async function upgradeGatewayActorDiamond(deployments) {
+    // Get the Gateway Diamond address from the deployments
+    const gatewayDiamondAddress = deployments.Gateway
```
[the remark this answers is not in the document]
snissn: I think we would need to pass the deployment.json around. It should just be public addresses! We need a lot of information to automatically do Diamond cut upgrades. We need to know the libs that the gateway and each facet was compiled with in order to determine the appropriate byte code on chain. At that point we are replicating the entire deployments json into command lines and it seems a lot easier to use the deployments json file!

scripts/util.ts
snissn: Hard hat compile generates it!

scripts/util.ts
snissn: Sure! Could I set that up in a follow on PR or do you want the readme blocking the merge?

adlrocha: It is fine if we add it to a follow-up PR, let me know if you want me to open an issue for this. Mainly, it would be really useful to document: 
* How upgrades work. 
* The process of performing a specific upgrades. 
* And the basic assets (like the `deployment.json`) that needs to be provided if we want someone else to perform the upgrade. 

Thanks! :pray:

adlrocha (approved):
Feel free to merge :) Just bear in mind to address my comment here: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/268#discussion_r1399359727 in a follow-up PR if possible. Thanks!

snissn:
follow up issue here: https://github.com/consensus-shipyard/ipc-monorepo/issues/126

PR #271, opened 2023-11-02
IPC-361: Pre-release and minor fixes

adlrocha:
This PR addresses the following issue raised by @aakoshh: https://github.com/consensus-shipyard/ipc/pull/361#pullrequestreview-1709252198
 
- It adds a `pre-release` method to the subnet contract to allow users to recover their initial balance if they need to before the subnet bootstraps.
- It fixes a bug in `leave` and allows validators leaving the subnet before it is bootstrapped to recover their initial balance.


src/lib/LibStaking.sol
```diff
@@ -438,7 +440,6 @@ library LibStaking {
         s.validatorSet.confirmWithdraw(validator, amount);
 
         // release stake from gateway and transfer to user
-        IGateway(s.ipcGatewayAddr).releaseStake(amount);
```
aakoshh, line 441: Let's see if I understand this: the stake is added to the gateway in one go when the subnet registers itself, right? So it makes sense that before that, this should not be trying to release anything. 

Strange we haven't caught any issues with this :thinking:

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -109,6 +109,31 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+
+        if (s.bootstrapped) {
+            revert SubnetAlreadyBootstrapped();
+        }
+
+        if (s.genesisBalance[msg.sender] == 0) {
+            revert NotEnoughBalance();
+        }
+
+        s.genesisBalance[msg.sender] -= amount;
```
aakoshh, line 128: I guess this can't go negative, right?

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -109,6 +109,31 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        s.genesisCircSupply -= amount;
+
+        if (s.genesisBalance[msg.sender] == 0) {
+            rmAddressFromBalanceKey(msg.sender);
+            payable(msg.sender).sendValue(amount);
```
aakoshh, line 133: But what happens if the balance is 5 and I try to withdraw 10. This seems to want to send 10 anyway.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -109,6 +109,31 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+    function preRelease(uint256 amount) external nonReentrant {
+        if (amount == 0) {
+            revert NotEnoughFunds();
+        }
+
+        if (s.bootstrapped) {
+            revert SubnetAlreadyBootstrapped();
+        }
+
+        if (s.genesisBalance[msg.sender] == 0) {
```
aakoshh, line 124:
```suggestion
        if (s.genesisBalance[msg.sender] < amount) {
```

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: It reverts on underflow/overflow

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: It will revert on 128 or 129

dnkolegov (approved):

src/lib/LibStaking.sol
adlrocha: I know. I feel that there were a bunch of tests that were commented so we lost track of this. But I included a few to catch it. Good that you raised this issue, as it already uncovered this bug :)

src/subnet/SubnetActorManagerFacet.sol
adlrocha: Good catch, I forgot to change it, this would've been a nasty bug. I will also include a unit test to catch this one. Thanks!

src/subnet/SubnetActorManagerFacet.sol
adlrocha: `payable` doesn't belong there, it goes out of that `if`, will be fixed in the next commit.

src/subnet/SubnetActorManagerFacet.sol
adlrocha: These are `uint256`, they can't be negative (they could theoretically underflow though). However, if `preFund` and `preRelease` don't have bugs this shouldn't happen.

In spite of this, do you think we should add a sanity-check just to be safe?

aakoshh: Ah, okay, I recommended the `balance < amount => revert` fix, but if it would revert even if you try an underflow, then it's okay, what matters is there is no way get more out.

aakoshh (approved):
LGTM!

PR #273, opened 2023-11-03
2 - Diamond extended upgrade support

snissn:
support for removing and adding facet functions (only function swaps was supported before)


adlrocha (approved):
This was a good catch. I didn't realize we couldn't delete facets. Quick question, are we covering in the contracts unit tests also these upgrade operations of removing facets? (I feel it would be useful to cover those in a follow-up PR if it is not currently the case).

snissn:
> This was a good catch. I didn't realize we couldn't delete facets in the previous PR. Quick question, are we covering in the contracts unit tests also these upgrade operations of removing facets? (I feel it would be useful to cover those in a follow-up PR if it is not currently the case). 

Yes we are covering removing facet functions in the tests! In both gateway and subnet actor tests

adlrocha (approved):
LGTM. If you have tested it and it works feel free to merge.

PR #274, opened 2023-11-03
run prettier on hardhat.confg.js 

snissn:
running Prettier on hardhat.config.js created a lot of big changes. 

 this commit only has the change from running prettier, so it is only code formating. 

this is done to make code review easier, and it is easy to verify for yourself that this is the only change


snissn:
i think this is creates way more confusion that otherwise it helps alleviate :)

PR #276, opened 2023-11-03
Check in ABI artifacts

aakoshh:
Unignore the `./out` directory so the step on CI which checks in the Rust bindings also checks in the JSON artifacts. 

With this change the Fendermint repo should no longer need to compile them.


adlrocha (commented):
Are we relying on the bindings pipeline to generate and commit this directory? I would really avoid relying on us (especially me :laughing:) for this. If this is the case, LGTM.

cc @cryptoAtwill and @dnkolegov to confirm too!

aakoshh:
That was the idea but I missed the fact that bindings won't always recompile - or maybe they would have but they only checks the `bindings` directory and they didn't in this case, so it didn't check in anything. I added a separate step to check `out`. I could probably have just checked it in manually and the `bindings` commit would have picked them up, but this should work too and less brittle.

aakoshh:
Almost 1 million lines :flushed:

PR #278, opened 2023-11-09
3 - enhance the subnet registry contract to follow diamond standard

snissn:


scripts/deploy-sa-diamond.ts
```diff
@@ -1,131 +1,147 @@
...
 
     type Libraries = {
-        [libraryName: string]: string;
+        [libraryName: string]: string
     }
 
     // ----
 
     const getterFacetLibs: Libraries = {
+        CheckpointHelper: libs['CheckpointHelper'],
```
adlrocha, line 34: I think we discussed it at some point (and this may be your linter introducing changes to the PR that are not relevant), but I think this file is outdated.

scripts/deploy-registry.template.ts
```diff
@@ -37,19 +37,70 @@ async function main() {
...
+      gateway: gatewayAddress,
+      getterFacet: getterFacet.address,
+      managerFacet: managerFacet.address,
+      subnetGetterSelectors: getterSelectors,
+      subnetManagerSelectors: managerSelectors,
+    };
+
+ 
+
+const facetCuts = [] //TODO
```
adlrocha, line 56: This is a typo, right? I see that you are indeed pushing the new facet cuts here.
```suggestion
const facetCuts = []
```

scripts/deploy-sa-diamond.ts
adlrocha: Oh, I see. It is because you included the right file extension, and the IDE linter picked it up, right?

src/SubnetRegistryDiamond.sol
```diff
@@ -0,0 +1,94 @@
...
+import {SubnetActorDiamond} from "./SubnetActorDiamond.sol";
+import {IDiamond} from "./interfaces/IDiamond.sol";
+ import {SubnetRegistryActorStorage} from "./lib/LibSubnetRegistryStorage.sol";
+ import {GatewayCannotBeZero, FacetCannotBeZero} from "./errors/IPCErrors.sol";
+ import {LibDiamond} from "./lib/LibDiamond.sol";
+error FunctionNotFound(bytes4 _functionSelector);
+
+contract SubnetRegistryDiamond {
+
+
```
adlrocha, line 13:
```suggestion
```

src/SubnetRegistryDiamond.sol
```diff
@@ -0,0 +1,94 @@
...
+    SubnetRegistryActorStorage internal s;
+
+
```
adlrocha, line 16: Are all these empty new lines on purpose?
```suggestion
```

adlrocha (approved):
Overall, LGTM. Thanks!

The only thing that we should bear in mind after merging is that we will need to update the `IpcProvider` to use the new implementation of the contract. Tracking this issue here: https://github.com/consensus-shipyard/ipc/issues/401

scripts/deploy-sa-diamond.ts
adlrocha: And I also see that this file has been brought up to date in this PR: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/280, thanks! (feel free to disregard this thread)

snissn: ty!! glad you connected the two! :)

src/SubnetRegistryDiamond.sol
snissn: The prettier commit down the line resolves these!  https://github.com/consensus-shipyard/ipc-solidity-actors/pull/283/files#diff-c17ce83c53d6d89c96379b0eec81c303f6bf66914ab439da0e1fe31769a03002L11

src/SubnetRegistryDiamond.sol
snissn: future commit fixes this

PR #284, opened 2023-11-13
Fix security scanner bugs

dnkolegov:
This PR fixes the bugs found by `4naly3er` scanner.


src/gateway/GatewayManagerFacet.sol
```diff
@@ -113,7 +113,8 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         if (!registered) {
             revert NotRegisteredSubnet();
         }
-        if (subnet.circSupply > 0) {
+        // gas-opt: original check: subnet.circSupply > 0
```
adlrocha, line 116: Context: this is possible because `circSupply` is an `uint256`, right?

adlrocha (approved):

src/gateway/GatewayManagerFacet.sol
dnkolegov: Correct

PR #285, opened 2023-11-20
Emit top down event

cryptoAtwill:
Emit top down messges are evens so that in ipc provider, there is no need to use getters. There are two benefits of doing this:
- No need to use getter and hence no need to perform look ahead for lotus nodes
- Save gas costs as storage of top down messages are shifted offchain instead of onchain. Topdown messages are not queried on chain anyway.


src/lib/LibGateway.sol
```diff
@@ -24,6 +24,9 @@ library LibGateway {
     using CheckpointHelper for BottomUpCheckpoint;
 
     event MembershipUpdated(Membership);
+    /// @dev nextSubnet refers to the next "down" subnet that the `CrossMsg.message.to` should be forwarded to.
+    /// @dev nonce is also indexed so that top down execution gas optimization can directly query messages by nonce.
+    event NewTopDownMessage(SubnetID indexed nextSubnet, uint64 indexed nonce, CrossMsg message);
```
aakoshh, line 29: Why call this `nextSubnet`?

src/lib/SubnetIDHelper.sol
```diff
@@ -118,10 +120,10 @@ library SubnetIDHelper {
     /// the subnet2 needs to be a subset of the subnet1
     function down(SubnetID calldata subnet1, SubnetID calldata subnet2) public pure returns (SubnetID memory) {
         if (subnet1.root != subnet2.root) {
-            return SubnetID({root: 0, route: new address[](0)});
+            revert DifferentRootNetwork();
         }
         if (subnet1.route.length <= subnet2.route.length) {
-            return SubnetID({root: 0, route: new address[](0)});
+            revert InvalidRoute();
         }
```
aakoshh, line 127: Looking at this and the docstring, I don't understand this check, or what the `down` function is trying to do. 

> the subnet2 needs to be a subset of the subnet1

What does _subset_ mean in this context? I reckon we should use _prefix_ and _suffix_ with these paths, they have obvious meanings. 

I would _think_ that `/a/b/c/d` is a "subset" of `/a/b`, but then I don't understand why its length should be shorter.

src/lib/SubnetIDHelper.sol
```diff
@@ -118,10 +120,10 @@ library SubnetIDHelper {
     /// the subnet2 needs to be a subset of the subnet1
     function down(SubnetID calldata subnet1, SubnetID calldata subnet2) public pure returns (SubnetID memory) {
         if (subnet1.root != subnet2.root) {
-            return SubnetID({root: 0, route: new address[](0)});
+            revert DifferentRootNetwork();
```
aakoshh, line 123: Maybe it's worth mentioning in the docstring that the return value changes from a value where the "unspecified" network show unrelated subnets to a `revert`. I'm not even sure you can call this a _pure_ function any more; a revert is pretty big side effect.

src/lib/LibGateway.sol
aakoshh: You could call it "child subnet" if you want to highlight that it's an immediate child of the current subnet. I'm wondering why you don't just use the address of the subnet actor?

src/lib/SubnetIDHelper.sol
cryptoAtwill: It's like the subnet to check is `/a/b/c/d` and the provided subnet/prefixes is `/a/b`, then the return should be `/a/b/c`

src/lib/SubnetIDHelper.sol
cryptoAtwill: I checked this function, it's only called in `topdown`. The thing with previous implementation is `SubnetID({root: 0, route: new address[](0)})` which is actually an error, it should not be returned to corrupt downstream processes.

src/lib/SubnetIDHelper.sol
aakoshh: Do you mean that `"/a/b/c/d".down("/a/b") == "/a/b/c"`? 

So, `/a/b` in this example is the _current subnet_ and `/a/b/c/d` is the _target subnet_, and `/a/b/c` is the _next subnet_. Renaming as such from `subnet1` and `subnet2` would be helpful, and using "prefix" instead of "subset" in the docs.

I can imagine that someone would want to call `down` to _check_ if a target is going up, or down. If it returned the _empty subnet_ then it did not want to go down; still not sure revert is the right thing to do but it's up to you guys.

src/lib/SubnetIDHelper.sol
aakoshh: Well yes,if the caller doesn't call the `isEmpty` function to check if they got a valid return value, then it would not go down well.

src/lib/LibGateway.sol
```diff
@@ -197,11 +200,13 @@ library LibGateway {
...
-        crossMessage.message.nonce = subnet.topDownNonce;
-        subnet.topDownNonce += 1;
+        uint64 topDownNonce = subnet.topDownNonce;
+
+        crossMessage.message.nonce = topDownNonce;
+        subnet.topDownNonce = topDownNonce + 1;
         subnet.circSupply += crossMessage.message.value;
 
-        s.topDownMsgs[subnetId.toHash()][block.number].push(crossMessage);
+        emit NewTopDownMessage({subnet: subnetId.toHash(), nonce: topDownNonce, message: crossMessage});
```
aakoshh, line 209: I'm looking at https://github.com/consensus-shipyard/ipc/blob/v0.6.0/ipc/sdk/src/subnet_id.rs and can't see an equivalent to whatever the `toHash` function is doing. 

Asking again: why don't use the address of the subnet actor the message is going to? This is a value I assume you need to be able to subscribe to, so the client has to be able to figure out what it should be, but it's even longer than the 20 byte address, and harder to get right.

src/lib/LibGateway.sol
```diff
@@ -24,6 +24,9 @@ library LibGateway {
     using CheckpointHelper for BottomUpCheckpoint;
 
     event MembershipUpdated(Membership);
+    /// @dev subnet refers to the next "down" subnet that the `CrossMsg.message.to` should be forwarded to.
+    /// @dev nonce is also indexed so that top down execution gas optimization can directly query messages by nonce.
+    event NewTopDownMessage(bytes32 indexed subnet, uint64 indexed nonce, CrossMsg message);
```
aakoshh, line 29: Is it really worth indexing the `nonce`? The filters only work on equality.

src/lib/LibGateway.sol
cryptoAtwill: `toHash` should be sth like: `ethers::utils::keccak256(ethers::abi::encode(&evm_subnet_id.into_tokens()))`

src/lib/LibGateway.sol
cryptoAtwill: This one I'm not that sure, but the idea is when we update top down message execution, we might need to execute events one by one and we need to query by nonce. This way will work if we cannot find particular nonce. But I guess you are right, we can remove it for now and add later since it's already upgradable.

src/lib/LibGateway.sol
aakoshh: If you go to a particular chain, the address is unique, it's the last item in the route. We do use that identifier for subnets. It's much easier to debug.

src/lib/LibGateway.sol
```diff
@@ -24,6 +24,8 @@ library LibGateway {
     using CheckpointHelper for BottomUpCheckpoint;
 
     event MembershipUpdated(Membership);
+    /// @dev subnet refers to the next "down" subnet that the `CrossMsg.message.to` should be forwarded to.
```
adlrocha, line 27:
```suggestion
    /// @dev `subnet` identifies the next subnet in the route of the message to its destination  (`CrossMsg.message.to`).
```

src/lib/SubnetIDHelper.sol
```diff
@@ -115,13 +117,15 @@ library SubnetIDHelper {
 
     /// @notice In the path determined by the current subnet id, it moves
     /// down in the path from the subnet id given as argument.
-    /// the subnet2 needs to be a subset of the subnet1
+    /// the subnet2 needs to be a subset of the subnet1.
```
adlrocha, line 120:
```suggestion
    /// subnet2 needs to be a prefix of the subnet1.
```

src/lib/SubnetIDHelper.sol
```diff
@@ -115,13 +117,15 @@ library SubnetIDHelper {
...
+    /// If subnet1 is /a/b/c/d and the subnet2  is /a/b, then the return should be /a/b/c.
```
adlrocha, line 121:
```suggestion
    /// If subnet1 is /a/b/c/d and subnet2 is /a/b, then the returned ID should be /a/b/c.
```

adlrocha (commented):
Would you mind adding a high-level description in the PR of how this approach prevents the look-ahead from top-down finality? 

I think the change makes sense (even if this doesn't fix parent finality), it improves the contract and simplifies the logic/storage (in exchange for a bit more of logic on the provider). But I would like to understand if this gives us something else I may be missing.

src/lib/LibGateway.sol
aakoshh: By the way for my understanding, did you change from a `SubnetIDd` to the hash because the compiler did not accept the former, or because you wanted to store less data? Asking because the [Topic](https://github.com/gakonst/ethers-rs/blob/dc6022e6152ec7ed33270eced6a061b087e5150e/ethers-core/src/types/filter.rs#L16) values are by design constrained to be U256 values; for `Address` they contain a copy of the 20 bytes padded to 32, and number are similar, but I assumed for larger stuff they might get hashed on their own, so maybe this hashing isn't needed at all? 

Also if we emit the event with the address, then it is actually trivial to just parse it and see where it's going without having to know what are the possible values, which would be nice for a display.

src/lib/LibGateway.sol
adlrocha: To determine if it is worth indexing by `nonce` it would be useful to understand how are you planning to trigger the message execution.

src/lib/LibGateway.sol
adlrocha: ^ agree. This is why I mentioned in my review that this maybe introducing additional non-negligible complexity in the provider :)

cryptoAtwill:
@adlrocha So that we can query events by block and we dont really have to use getter as getter needs height + 1 finalization.

adlrocha (approved):
Ok, now I get it. So the high-level idea is the following: 
* Before we were indexing top-down messages by height, which meant that due to the deferred execution used in Lotus, in order to get the messages to be executed for height `n`, we had to wait for height `n+1`, but handling this with null rounds was leading to inconsistencies between validators. 
* With this change, we emit an event when top-down messages are committed in a block, so the messages to be executed for block `n` are immediately available in the block without requiring an additional query to the state (preventing the height dance that we needed before as there was one height delay between the block being available and the state computation).  Additionally, this approach is cleaner and gas efficient.

src/lib/LibGateway.sol
cryptoAtwill: I changed from `SubnetID` to hash because evm will hash it anyways. Yeah, if everyone is ok with address, I can change to address.

cryptoAtwill:
@adlrocha @aakoshh, I changed the index to address, if you want to take a look, else I will merge this PR.

cryptoAtwill:
@adlrocha @aakoshh, I changed the index to address, if you want to take a look, else I will merge this PR.

adlrocha (approved):
Yes, I think indexing by address is easier than using a hash. This LGTM, thanks!

src/lib/LibGateway.sol
aakoshh: I think the idea is to use it exactly like how the [validator updates](https://github.com/consensus-shipyard/ipc/blob/3c2a112e35c6592b3eb4dbf57724bb5900493936/ipc/provider/src/manager/evm/manager.rs#L177-L180) are fetched, using the `eth_getLogs` RPC method.

aakoshh:
> With this change, we emit an event when top-down messages are committed in a block, so the messages to be executed for block n are **immediately available** in the block without requiring an additional query to the state

(emphasis mine)

Note that this isn't true, we still have to execute the block, and therefore we have to wait for block N+1 to be available in the parent before we can query this thing, otherwise it can return empty list.

And I don't think a query such as [this one](https://github.com/consensus-shipyard/ipc/blob/3c2a112e35c6592b3eb4dbf57724bb5900493936/ipc/provider/src/manager/evm/manager.rs#L198C57-L198C57) actually protects against this, because block N can be available, but unexecuted.

So, overall, I think what I said in https://github.com/consensus-shipyard/ipc-monorepo/issues/158 still applies: you have to find the last height you are guaranteed to have executed when block N is final, and query that one.

adlrocha:
> Outcome of the Slack discussion on this matter: https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1700562534384499

In the previous implementation:
* We saw block N as final through the RPC API.
* In order to request the top-down messages committed at height N we had to wait for the state to be available at height N+1 to query the state, and this was a mess because it wasn't an atomic operation.
Now, by emitting and querying events we made this operation atomic.
* We see block N and the events are available as part of the receipt of `N`.
* We don't need to explicitly inspect the state at height N+1 , because what we query is the receipt for block N.

However, if we look at just [get_validator_changeset](https://github.com/consensus-shipyard/ipc/blob/3c2a112e35c6592b3eb4dbf57724bb5900493936/ipc/provider/src/manager/evm/manager.rs#L164-L204) it does query a specific height but doesn't wait to see if the receipt was already executed, which may lead to a validator seeing no events when actually this is not the case and it just a matter of the receipt not being executed yet.

As @aakoshh suggests, the best way to tackle this is to: 
* look back and find the last non-null height, and get the events up to that. (edited) 
* change this:
```
                    let validator_changes = provider
                        .validator_changes_from(prev_height + 1, finality.height)
```
into something like:
```
    let last_non_null_height = seek_backwards_from(finality.height-1);
    let validator_changes = provider
                        .validator_changes_from(prev_height, last_non_null_height)
```
So when parent finality N is committed, I must get the receipts for [latest_finality, N-1] and execute the side-effects, that way we can be sure that all receipts are available. This applies to validator_set and top-down messages.

This requires no changes in the current implementation of the contract in this PR, but it requires changes in the way Fendermint commits the parent finality.

PR #293, opened 2023-11-24
Contract upgrade tooling: makefile + documentation

snissn:


hardhat.config.ts
```diff
@@ -190,27 +190,6 @@ task(
     },
 )
 
-task(
```
adlrocha, line 193: Can you double-check that I didn't mess up when rebasing `dev`? Did you mean to remove this and the `deploy-subnet.ts` file?

README.md
```diff
@@ -49,6 +49,48 @@ To deploy the contracts in some other network configured in the Hardhat config y
...
 
+# Upgrading IPC Solidity Contracts
+
+This repository's contracts use the Diamond pattern for upgradability, allowing new features to be added or issues to be corrected without a full redeployment. The upgrade process is automated and includes bytecode verification to ensure the integrity of the changes.
+
+## Automated Upgrade and Bytecode Verification
+
+When you run an upgrade command, the repository's scripts handle several tasks:
+
+1. **Bytecode Verification**: The scripts fetch the bytecode of the currently deployed contracts from the FEVM network using the details stored in local JSON files. They compare this with the bytecode generated after applying the intended changes on a temporary Ganache network.
```
adlrocha, line 60: Would you mind adding a note pointing to the default directory where the deployment assets are placed? These assets will be the ones that we will need to persist to allow anyone from the team to perform upgrades.

README.md
```diff
@@ -49,6 +49,48 @@ To deploy the contracts in some other network configured in the Hardhat config y
...
+-   **Gateway Diamond Upgrade**:
+
+    ```bash
+    make upgrade-gw-diamond [NETWORK=<network-name>]
+    ```
+
+-   **Subnet Actor Diamond Upgrade**:
+
+    ```bash
+    make upgrade-sa-diamond [NETWORK=<network-name>]
```
adlrocha, line 77: Can a subnet actor be upgraded if it wasn't initially deployed using hardhat but the `ipc-cli`? This would mean that the deployment assets are not available for deployment.

hardhat.config.ts
snissn: correct as far as i can tell, those were unmaintained somewhat duplicated commands that i cleaned up in this PR

README.md
snissn: I don't know enough about ipc-cli to answer this question.. does ipc-cli end up calling hardhat?

README.md
snissn: #1bab656

README.md
adlrocha: Unfortunately not, it makes a transaction to the registry to trigger the deployment of a new subnet actor. Is there a way to artificially generate a `desployment.json` to help with the upgrade? We may need to come up with a way to allow updates from subnets deployed from the registry.

README.md
```diff
@@ -49,6 +49,48 @@ To deploy the contracts in some other network configured in the Hardhat config y
...
 
+# Upgrading IPC Solidity Contracts
+
+This repository's contracts use the Diamond pattern for upgradability, allowing new features to be added or issues to be corrected without a full redeployment. The upgrade process is automated and includes bytecode verification to ensure the integrity of the changes.
+
+## Automated Upgrade and Bytecode Verification
+
+When you run an upgrade command, the repository's scripts handle several tasks:
+
+1. **Bytecode Verification**: The scripts fetch the bytecode of the currently deployed contracts from the FEVM network using the details stored in local JSON files in the root directory of the github repository. They compare this with the bytecode generated after applying the intended changes on a temporary Ganache network.
```
maciejwitowski, line 60: Should we replace "FEVM network" with "IPC parent network"? IIUC these actors aren't aware of FEVM/Filecoin etc, right?

README.md
snissn: I think I understand better how this code works and what needs to happen for this to work from doing the internal audit. I think that there's a reasonable way through this where we need to add code in ./src/subnetregistry/RegisterSubnetFacet.sol that includes the diamond cut facets and then we will just need to exapnd the facets on the subnet registry to manage the upgrades. we'll need some sort of management for the equivalent of deployment.json but that might not be so bad and it might make sense to store some of these data items on chain in the registry. One last concern is that we may end up with a system that has too many subnets to upgrade in a single block so we should have a fallback there to paginate and may need to think about whether we need to do this upgrade change in a transaction where contracts are paused during the upgarde if it spans multiple txs

adlrocha:
hey @snissn. Do you mind rebasing this one and fixing the conflicts? I think it may be a good idea to get this one merged and address my comment about the subnet upgrade mechanism in a follow-up PR (in order not to leave this one hanging too long).

README.md
raulk: Now tracking here: https://github.com/consensus-shipyard/ipc-monorepo/issues/497

README.md
raulk:
```suggestion
1. **Bytecode Verification**: The scripts fetch the bytecode of the currently deployed contracts on an FEVM-powered IPC network using the details stored in local JSON files in the root directory of the git repository. They compare this with the bytecode generated after applying the intended changes on a temporary Ganache network.
```

raulk (approved):
There seems to be a bunch of improvements still needed here, but let's merge this and continue working on the rest in the ipc-monorepo.

PR #296, opened 2023-11-28
gateway: fund/release: revert when zero value is sent.

raulk:
This prevents xnet message spamming, especially in the fund path, which is exempt of fees on the other side. We should revisit the free-of-charge nature here since it can lead to other forms of spam.

TODO adapt unit tests.


src/gateway/GatewayManagerFacet.sol
```diff
@@ -127,10 +127,18 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
...
+    /// There may be an associated fee that gets distributed to validators in the subnet. Currently this fee is zero,
+    /// i.e. funding a subnet is free.
+    ///
+    /// @param subnetId: the destination subnet for the funds.
+    /// @param to: the address to which to credit funds in the destination subnet.
     function fund(SubnetID calldata subnetId, FvmAddress calldata to) external payable {
+        if (msg.value == 0) {
+            // prevent spamming if there's no value to fund.
+            revert InvalidCrossMsgValue();
+        }
```
raulk, line 141: Can also be refactored as Solidity modifier / guard.

PR #297, opened 2023-11-29
Fix prettier to not inspect package.json

dnkolegov:
This PR makes prettier to ignore `package.json` and `package-lock.json`


.github/workflows/prettier.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Run Prettier on Solidity Files
         run: npx prettier --write '**/*.sol'
       - name: Debugging - Show changed files
-        run: git status && git diff --name-only
+        run: git restore package-lock.json && git status && git diff --name-only
```
adlrocha, line 33: Do we really need the `git restore` if we are including to the `.prettierignore`?

adlrocha (approved):
A minor comment but this makes sense. Thank you!

.github/workflows/prettier.yml
dnkolegov: Yes, we need it. Because that workflow installs prettier and rewrites the `package-lock.json`

PR #300, opened 2023-11-30
Add a basic invariant test for registry

dnkolegov:
This PR adds an invariant test for the registry contract.

### DIFF
--- a/test/mocks/SubnetActor.sol
+++ b/test/mocks/SubnetActor.sol
@@ -4,7 +4,7 @@ pragma solidity 0.8.19;
 import {SubnetActorManagerFacet} from "../../src/subnet/SubnetActorManagerFacet.sol";
 import {LibStaking} from "../../src/lib/LibStaking.sol";
 
-contract SubnetManagerTestUtil is SubnetActorManagerFacet {
+contract SubnetActorManagerFacetMock is SubnetActorManagerFacet {
     function confirmChange(uint64 _configurationNumber) external {
         LibStaking.confirmChange(_configurationNumber);
     }


test/handlers/SubnetRegistryHandler.sol
```diff
@@ -0,0 +1,136 @@
...
+        if (_minCollateral > DEFAULT_MIN_VALIDATOR_STAKE || _minCollateral == 0) {
+            _minCollateral = DEFAULT_MIN_VALIDATOR_STAKE;
+        }
+        if (_bottomUpCheckPeriod > DEFAULT_CHECKPOINT_PERIOD || _bottomUpCheckPeriod == 0) {
+            _bottomUpCheckPeriod = DEFAULT_CHECKPOINT_PERIOD;
+        }
+        if (_majorityPercentage < 51 || _majorityPercentage > 100) {
+            _majorityPercentage = DEFAULT_MAJORITY_PERCENTAGE;
+        }
+        if (_powerScale > DEFAULT_POWER_SCALE) {
```
adlrocha, line 97: powerScale can also be negative

test/handlers/SubnetActorHandler.sol
```diff
@@ -0,0 +1,152 @@
...
+contract SubnetActorHandler is CommonBase, StdCheats, StdUtils {
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    SubnetActorManagerFacetMock private managerFacet;
+    SubnetActorGetterFacet private getterFacet;
+
+    uint256 private constant DEFAULT_MIN_VALIDATOR_STAKE = 1 ether;
+    uint64 private _configurationNumber = 1;
+
+    // Ghost variables.
```
adlrocha, line 33: Out of a curiosity, why are they called ghost variables?

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -15,6 +15,8 @@ import {LibValidatorSet, LibStaking} from "../lib/LibStaking.sol";
 import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
 import {Address} from "openzeppelin-contracts/utils/Address.sol";
 
+uint256 constant ETH_SUPPLY = 120_500_000 ether;
```
adlrocha, line 18: What is this `ETH_SUPPLY`?  It is the total expected circulating supply of ETH in Ethereum? If this is the case, we shouldn't use this number. Subnets may have their own total circulating supply.

test/SubnetActorInvariants.t.sol
```diff
@@ -0,0 +1,311 @@
...
+        targetContract(address(subnetActorHandler));
+    }
+
+    /// @notice The number of joined validators is equal to the number of total validators.
+    function invariant_SA_01_validators_number_is_correct() public {
+        assertEq(saGetter.getTotalValidators(), subnetActorHandler.joinedValidatorsNumber());
+    }
+
+    /// @notice The sum of the Subnet Actor Handler's ETH balance plus the total staked sum should equal the total ETH_SUPPLY.
+    function invariant_SA_02_conservationOfETH() public {
```
adlrocha, line 144: I really don't understand this invariant, you are checking that the total collateral send by validators is equal to the subnet actors balance? This will only be the case if the gateway hasn't registered yet. When this happens the collateral is delegated to the gateway.

test/handlers/SubnetRegistryHandler.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    function getOwners() external view returns (address[] memory) {
+        return ghost_owners.values();
+    }
+
+    function getGateway() external view returns (address) {
+        return registerGetterFacet.getGateway();
+    }
+
+    function deploySubnetActor(
```
adlrocha, line 77: Not a strong opinion, but when I saw this `deploySubnetActor` function I was like "wait, didn't we have a function already that deployed the contract before?", until I realized that this triggers the deployment from the registry.
```suggestion
    function deploySubnetActorFromRegistry(
```

adlrocha (commented):
I left a few comments and requests for clarification. They shouldn't be blocking but just to double-check that we are considering all of the scenarios. Thanks!

specs/PROPERTIES.md
```diff
@@ -0,0 +1,30 @@
...
+| -------- | ---------------------------------------------------------- | -------------------- | ------ |
+| SR-01    | The Gateway address is not changed                         | Variable Transitions | ✅     |
+| SR-02    | If a subnet was created then it's address can be retrieved | High Level           | ✅     |
+
+## Subnet Actor
+
+| Property | Description                                                                                                    | Category             | Tested |
+| -------- | -------------------------------------------------------------------------------------------------------------- | -------------------- | ------ |
+| SA-01    | The number of joined validators is equal to the number of total validators.                                    | Variable Transitions | ✅     |
+| SA-02    | The sum of the Subnet Actor Handler's ETH balance plus the total staked sum should equal the total ETH_SUPPLY. | Unit Test            | ✅     |
```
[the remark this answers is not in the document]
dnkolegov: Handler is a contract that implements an assumptions and interactions model with the target contract. It is a term from invariant testing.
We do not fuzz actors directly, we fuzz handler that calls target contracts.

test/handlers/SubnetActorHandler.sol
dnkolegov: That is another term from invariant tests. Ghost variables are variables that live in the handler contracts and are used to track sent/received values, count calls, etc

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: Sorry, that is the wrong contract. It should be in the handler.

dnkolegov: Fixed

test/handlers/SubnetRegistryHandler.sol
dnkolegov: Fixed

specs/PROPERTIES.md
```diff
@@ -0,0 +1,30 @@
...
+| SA-03    | The sum of the validator stakes is equal to the total confirmed collateral.                                    | Valid State          | ✅     |
+| SA-04    | After leving the subnet, a validator can withdraw all ETHs that were staked.                                   | High Level           | ✅     |
```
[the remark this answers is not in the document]
dnkolegov: Fixed

specs/PROPERTIES.md
```diff
@@ -0,0 +1,30 @@
...
+| SR-01    | The Gateway address is not changed                         | Variable Transitions | ✅     |
+| SR-02    | If a subnet was created then it's address can be retrieved | High Level           | ✅     |
+
+## Subnet Actor
+
+| Property | Description                                                                                                    | Category             | Tested |
+| -------- | -------------------------------------------------------------------------------------------------------------- | -------------------- | ------ |
+| SA-01    | The number of joined validators is equal to the number of total validators.                                    | Variable Transitions | ✅     |
+| SA-02    | The sum of the Subnet Actor Handler's ETH balance plus the total staked sum should equal the total ETH_SUPPLY. | Unit Test            | ✅     |
+| SA-03    | The sum of the validator stakes is equal to the total confirmed collateral.                                    | Valid State          | ✅     |
```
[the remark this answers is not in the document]
dnkolegov: The current setup suggests that the subnet is bootstraped and all changes are confirmed immediately via mock.

test/handlers/SubnetRegistryHandler.sol
dnkolegov: They can, here they have an upper bound.

adlrocha (approved):

PR #305, opened 2023-12-04
IPC-304: add permissioned flag to disable validator changes

### LINKED ISSUE #304: Disable validator changes after subnet is bootstrapped
We discussed removing the ability of triggering any validator changes after a subnet is bootstrapped as a first stage towards permissioned/federated subnets.

adlrocha:
Fixes #304 

## Background
This is a first stage towards adding the option to make subnets permissioned. In this case, I included a `permissioned` flag that can be set in construction that determines if validator changes should be disabled or not.

Why using a construction flag? Ideally, I would've loved to use a compilation flag, so you can already compile a contract with the subset of features that you want, but Solidity doesn't have support for compilation flags, and there doesn't seem to be an easy way of getting an analogous behavior without a lot of redundant code. This approach also allows us to have this functionality without having to explicitly re-write or break any of our existing code. 

In the short-term, I expect validator changes to be disable for Mycelium and other public deployments, while in user child subnets we may want to leave them enabled so users can test the full-functionality of testnets. 

## Next steps
I wanted to double-check that everyone is OK with this approach before making further progress. If this is the case, the next steps is to: 
* Add a few unit tests covering the `permissioned` scenario. 
* Propagate the changes to the contract interface and storage to `ipc` and `fendermint` (I will open the corresponding issues).

### DIFF
--- a/src/SubnetActorDiamond.sol
+++ b/src/SubnetActorDiamond.sol
@@ -31,6 +31,7 @@ contract SubnetActorDiamond {
         uint16 activeValidatorsLimit;
         uint256 minCrossMsgFee;
         int8 powerScale;
+        bool permissioned;
     }
 
     constructor(IDiamond.FacetCut[] memory _diamondCut, ConstructorParams memory params) {
@@ -70,6 +71,7 @@ contract SubnetActorDiamond {
         s.powerScale = params.powerScale;
         s.minCrossMsgFee = params.minCrossMsgFee;
         s.currentSubnetHash = s.parentId.createSubnetId(address(this)).toHash();
+        s.permissioned = params.permissioned;
 
         s.validatorSet.activeLimit = params.activeValidatorsLimit;
         // Start the next configuration number from 1, 0 is reserved for no change and the genesis membership
--- a/src/subnet/SubnetActorManagerFacet.sol
+++ b/src/subnet/SubnetActorManagerFacet.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity 0.8.19;
 
-import {SubnetAlreadyBootstrapped, NotEnoughFunds, CollateralIsZero, CannotReleaseZero, NotOwnerOfPublicKey, EmptyAddress, NotEnoughBalance, NotEnoughBalanceForRewards, NotEnoughCollateral, NotValidator, NotAllValidatorsHaveLeft, NotStakedBefore, InvalidSignatureErr, InvalidCheckpointEpoch, InvalidCheckpointMessagesHash, InvalidPublicKeyLength} from "../errors/IPCErrors.sol";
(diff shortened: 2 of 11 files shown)


jsoares:
Fwiw, I don't think this is the sort of thing that would be much better implemented as a compilation flag. It'd be great to have support for those _generally_, but this approach looks fine.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -223,8 +239,16 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     /// @dev it also return the validators initial balance if the
     /// subnet was not yet bootstrapped.
     function leave() external notKilled nonReentrant {
-        // remove bootstrap nodes added by this validator
+        // disbling validator changes for permissioned subnets (at least for now
+        // until a more complex mechanism is implemented).
+        // This means that initial validators won't be able to recover
+        // their collateral ever (worth noting in the docs if this ends
+        // up sticking around for a while).
+        if (s.permissioned) {
```
maciejwitowski, line 247: 1) Probably I missed the discussion but why are we blocking `leave` as well? 
2) If we decide to do that, shouldn't we enforce it only for the already boostrapped network to keep it symmetrical with `join` (maybe not important in practice but easier to reason about IMO)

src/SubnetActorDiamond.sol
```diff
@@ -31,6 +31,7 @@ contract SubnetActorDiamond {
         uint16 activeValidatorsLimit;
         uint256 minCrossMsgFee;
         int8 powerScale;
+        bool permissioned;
```
dnkolegov, line 34: Technically, it is not permissioned. Here we do not allow to join even if the validator is "permissioned". I do not the correct term for that.

jsoares: `static` is probably the accurate term, but not too descriptive. `staticMembership` ?

dnkolegov: Yes, `static` sounds good

adlrocha: The reason why I went with `permissioned` is because this is a first stage towards this design for federated validation: https://www.notion.so/pl-strflt/Federated-validation-6290874b58904818ac4fdddb5128d01b 

I have no strong opinions, I can name this `static` until we have the `permissioned` feature.

src/subnet/SubnetActorManagerFacet.sol
jsoares: If we're limiting membership changes to reduce the changes of things going wrong, it makes sense to limit them both ways and just not have to worry about those paths. Don't see a lot of value in keeping leave as presumably you're not starting with a large validator set that you can afford to lose over time.

adlrocha: > Probably I missed the discussion but why are we blocking leave as well?

We don´t  want any validator change for now, as this is what is breaking Mycelium :)

> If we decide to do that, shouldn't we enforce it only for the already boostrapped network to keep it symmetrical with join (maybe not important in practice but easier to reason about IMO)

This is a good point, fixed it

maciejwitowski (approved):

PR #309, opened 2023-12-08
Federated Validation

cryptoAtwill:
Adding federation validation. The approach is as follows:
- Add a `setMembership` function that takes a list of validators and their power
- Update LibStaking with `setFederatedPower` method that pushes the change to queue
- Add `confirmFederatedPower` that confirms the change
- Sort validators based on total power, power is calculated with `federatedPower + collateral`


src/lib/LibStaking.sol
```diff
@@ -151,23 +162,23 @@ library LibValidatorSet {
 
     /// @notice Get the confirmed collaterals of the validators.
```
dnkolegov, line 163: The description is outdated: total vs confirmed

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -133,6 +141,21 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+    function setFederatedPowers(address[] calldata validators, uint256[] calldata powers) external onlyOwner notKilled {
+        if (validators.length != powers.length) {
+            revert InvalidFederationPayload();
+        }
+
+        for (uint i = 0; i < validators.length; ) {
+            LibStaking.setFederatedPower(validators[i], powers[i]);
+
+            unchecked {
+                i++;
```
dnkolegov, line 154: Is it intentional? Why not `++i`?

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -133,6 +141,21 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         payable(msg.sender).sendValue(amount);
     }
 
+    /// @notice method that allows the contract owner to set the validators' federated power
+    function setFederatedPowers(address[] calldata validators, uint256[] calldata powers) external onlyOwner notKilled {
+        if (validators.length != powers.length) {
+            revert InvalidFederationPayload();
+        }
+
+        for (uint i = 0; i < validators.length; ) {
```
dnkolegov, line 150: Please use `for (uint256 i; i < validators.length; ) {`

dnkolegov (changes requested):
- Added several minor things that should be fixed
- The main thing is I do not see any tests here despite the fact you are adding a new feature.

maciejwitowski:
> Add a setMembership function that takes a list of validators and their power

Hm, I can't find `setMembership` in this PR. Am I missing something?

src/lib/LibStaking.sol
```diff
@@ -107,6 +107,17 @@ library LibValidatorSet {
...
 
+    /// @notice Get the total voting power for the validator
+    function getPower(
+        ValidatorSet storage validators,
+        address validator
+    ) internal view returns(uint256 power) {
+        uint256 collateral = validators.validators[validator].confirmedCollateral;
+        uint256 federatedPower = validators.validators[validator].confirmedCollateral;
+
+        power = collateral + federatedPower;
```
maciejwitowski, line 118: One of the requirements in the design doc says
> The power of validators is not determined by their collateral in the system

Which looks inconsistent with this line (increasing `federatedPower`-based power with `collateral` ).

cryptoAtwill: so the power is purely determined by `federatedPower`? Then what's the stack for?

cryptoAtwill:
@maciejwitowski It's called `setFederatedPower`.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -26,6 +27,13 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     event NextBottomUpCheckpointExecuted(uint64 epoch, address submitter);
     event SubnetBootstrapped(Validator[]);
 
+    modifier onlyOwner() {
```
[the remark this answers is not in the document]
cryptoAtwill: I didnt know there is a way to call modifier from a `lib`, is it possible?

src/subnet/SubnetActorManagerFacet.sol
cryptoAtwill: is there a difference? Or it's a preference?

cryptoAtwill:
@dnkolegov Please see the below:

> * Added several minor things that should be fixed

I have updated them.

> * The main thing is I do not see any tests here despite the fact you are adding a new feature.

It's an initial impl and just check in my understanding, will add them once confirmed implementation so that I dont have to change tests.

src/subnet/SubnetActorManagerFacet.sol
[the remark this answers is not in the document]
dnkolegov: Do we not do the same in LibReentrancyGuard?

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: There is. They work differently. `++i` is cheaper. If there is no need to use `i++` we use `++i` everywhere.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -147,11 +147,11 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
             revert InvalidFederationPayload();
         }
 
-        for (uint i = 0; i < validators.length; ) {
+        for (uint256 i = 0; i < validators.length; ) {
```
dnkolegov, line 150: You should also delete `0` from the variable declaration. 
And store the length of the array outside of the loop.

`for (uint256 i; i < n; ) {`

dnkolegov (changes requested):

src/subnet/SubnetActorManagerFacet.sol
cryptoAtwill: oh I see what you mean. Yep, I will update.

src/lib/LibStaking.sol
maciejwitowski: cc @adlrocha

adlrocha (changes requested):
I feel you ended up something more complicated than what we originally wanted. As described in my comments we want to have either a power table defined by collateral, or by power weight given by `setMembership` (what you are calling `federatedPower`). The mechanics of proposing changes and confirming should stay the same.

Ideally, we should also maybe put this permissioning into its own Facet to decouple it from the main logic and make it easy to replace.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -133,6 +141,21 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
     }
 
+    /// @notice method that allows the contract owner to set the validators' federated power
+    function setFederatedPowers(address[] calldata validators, uint256[] calldata powers) external onlyOwner notKilled {
+        if (validators.length != powers.length) {
+            revert InvalidFederationPayload();
+        }
+
+        for (uint256 i = 0; i < validators.length; ) {
+            LibStaking.setFederatedPower(validators[i], powers[i]);
```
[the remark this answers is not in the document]
cryptoAtwill: that means if `permissioned == true`, we should also disable join/leave?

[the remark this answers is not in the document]
cryptoAtwill: my question is do we still do:
1. Order validators by their power, i.e. rank them by their power and active vs non-active?
2. Do we need validators to confirm the changes?
3. No collateral required for the federated validators?

How do we initialized fendermint from `genesis`? Will exiting scripts for generating genesis break?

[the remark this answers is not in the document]
cryptoAtwill: If we dont do this, then we will break the entire `fendermint` validator update flow or at least requires some major changes.

test/integration/SubnetActorDiamond.t.sol
```diff
@@ -1072,4 +1073,109 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
     function callback() public view {
         // console.log("callback called");
     }
+
+    // function _assertJoin(address validator, uint256 amount) internal {
```
adlrocha, line 1077: Why are we commenting this tests? If this is because there is no easy way to deploy a gateway with other parameters we should parametrize the test setup so we can cover all possible tests and different situations.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -133,6 +141,40 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
         payable(msg.sender).sendValue(amount);
     }
 
+    /// @notice method that allows the contract owner to set the validators' federated power
+    function setFederatedPowers(
```
adlrocha, line 145: I am honestly not a fan of this name for the function, we could maybe call it `setMembershipConfig` which is the more "academic" name for what we are doing? 
```suggestion
    function setFederatedPower(
```

src/structs/Subnet.sol
```diff
@@ -91,6 +94,14 @@ struct ValidatorInfo {
     bytes metadata;
 }
 
+/// Determines the permission mode for validators.
+enum PermissionMode {
```
adlrocha, line 98: I think it would be cleaner if you remove the `permissioned` flag and add here a `Static` permission mode to flag where no changes to the validator set is allowed.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -133,6 +141,40 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+        address[] calldata validators,
+        bytes[] calldata publicKeys,
+        uint256[] calldata powers
+    ) external onlyOwner notKilled {
+        if (!s.permissioned) {
```
adlrocha, line 150: As described in a comment above, ideally we should check the permission mode here to determine if the method is allowed or not.

src/SubnetActorDiamond.sol
```diff
@@ -73,6 +73,11 @@ contract SubnetActorDiamond {
         s.currentSubnetHash = s.parentId.createSubnetId(address(this)).toHash();
         s.permissioned = params.permissioned;
 
+        if (params.permissioned) {
```
adlrocha, line 76: I think it may be good to pass the `permissionedMode` here instead of using a flag, maybe?

src/lib/LibStaking.sol
```diff
@@ -138,36 +150,36 @@ library LibValidatorSet {
...
+    /// @notice Get the total power of the validators.
     /// The function reverts if at least one validator is not in the active validator set.
-    function getConfirmedCollaterals(
+    function getTotalPowerOfValidators(
         ValidatorSet storage validators,
         address[] memory addresses
     ) internal view returns (uint256[] memory) {
         uint256 size = addresses.length;
-        uint256[] memory activeCollaterals = new uint256[](size);
+        uint256[] memory activePowers = new uint256[](size);
```
adlrocha, line 171: Instead of using `Powers` should we maybe use `PowerTable` throughout the code? I think way is clearer what we mean.   
```suggestion
        uint256[] memory activePowerTable = new uint256[](size);
```

src/structs/Subnet.sol
```diff
@@ -83,6 +84,8 @@ struct StakingReleaseQueue {
 ///     - Confirmed: The amount of collateral actually confirmed in child subnet
 ///     - Total: Aside from Confirmed, there is also the collateral has been supplied, but not yet confirmed in child.
 struct ValidatorInfo {
+    /// The power set by contract admin
+    uint256 federatedPower;
```
adlrocha, line 88: I understand that it is simpler this way, but as we discussed sync I am not a fan of having `federatedPower` and `collateralPower` as to distinct entities. Ideally `ValidatorInfo` should just have a `power` field and according to the `PermissionMode` the specific logic is triggered. I feel this would allow us in the future to abstract and modularize the permissioning system.  But I guess that with the time we have left it may be good enough :shrug:

adlrocha (commented):
This is looking better. A few suggestions before we approve and merge:
* Let's remove the `permissioned` flag and just use `permissionMode` to determine the permissioning logic used, it will help abstracting this in the future. 
* As I've mentioned a few times, I am not a fan of having `federatedPower` and `collateralPower` as two different fields, I feel we should just have `power` and decouple the permissioning logic so it decides how to interpret the power and how it should be updated. I understand that making this change may require some refactoring and it would take more time, but it would help if we want to fully abstract this in the future. In any case, this should work, so I will let @maciejwitowski have the final call on this, considering the time constraints :)  (also, Maciej, related to our huddle, this is a good example of the trade-offs between doing the right thing or moving fast that we've been having).

src/structs/Subnet.sol
cryptoAtwill: Yeah, but if we change to power, a lot of the parameters and their meanings need to be updated as well. Like in the release queue, we are allowing validators to withdraw their collateral, which makes total sense. But if we change to power and interpret as collateral, it's withdrawing power, which I feel will be misleading.

src/SubnetActorDiamond.sol
cryptoAtwill: It would break existing `fendermint` and `ipc` code base, or do we code freeze there as well?

maciejwitowski:
@adlrocha @cryptoAtwill IIUC we are in the situation where we can either go into Solidity audits with the current implementation (+ fixing all PR comments ofc) or none. I agree we should refactor but I doubt we'd land it by tomorrow. 
In this case, I think we should stick with the current implementation + adding more tests.

adlrocha:
Wait, why has this PR been closed? Fixing all my comments except https://github.com/consensus-shipyard/ipc-solidity-actors/pull/309#discussion_r1425292767 (that we chose to defer), should make it good to go?

cryptoAtwill:
@adlrocha I closed it temporarily because it keeps on triggering the CI which creates a lot of rebase and uses unnecessary resources. I tried marking it as draft, but still runs.

src/subnet/SubnetActorManagerFacet.sol
cryptoAtwill: I have updated them all.

src/structs/Subnet.sol
cryptoAtwill: That flag is removed in storage, but I kept it in the contructor params as this way it wont break existing `fendermint` and `ipc` codebase.

binding/src/lib_gateway.rs
```diff
@@ -132,13 +132,21 @@ pub mod lib_gateway {
         __abi,
     );
     #[rustfmt::skip]
+<<<<<<< HEAD
```
maciejwitowski, line 135: There are bunch of unaddressed merge conflicts. Not sure what the plan is as I see this is active topic in the Slack thread.

src/SubnetActorDiamond.sol
maciejwitowski: We don't code freeze `fendermint` and `ipc` yet (today). IIUC we could use the `permissionedMode` and adapt other repos, right?

maciejwitowski (commented):
I guess @adlrocha may want to take another look but I did a pass since the time is limited today.

2 things
- The only unaddressed comment is about [function name change](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/309/files#r1425273419). It's a small thing but let's address for completeness 
- There are not fixed merge commits in rust bindings. These have to be fixed (depending on the slack convo outcome)

Other than that LGTM.

maciejwitowski (approved):

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -186,11 +235,10 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
 
     /// @notice method that allows a validator to increase its stake
     function stake() external payable notKilled {
-        // disbling validator changes for permissioned subnets (at least for now
+        // disbling validator changes for federated validation subnets (at least for now
         // until a more complex mechanism is implemented).
-        if (s.permissioned) {
```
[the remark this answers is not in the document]
cryptoAtwill: that means we should `enforceCollateralValidation` right? The implementation does that?

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -133,14 +148,48 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
+
     /// @notice method that allows a validator to join the subnet
     /// @param publicKey The off-chain 65 byte public key that should be associated with the validator
     function join(bytes calldata publicKey) external payable nonReentrant notKilled {
         // adding this check to prevent new validators from joining
         // after the subnet has been bootstrapped. We will increase the
         // functionality in the future to support explicit permissioning.
-        if (s.bootstrapped && s.permissioned) {
-            revert MethodNotAllowed();
+        if (s.bootstrapped) {
```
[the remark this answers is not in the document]
cryptoAtwill: The current implementation still allows join after bootstrap, there is a test case covering it.

adlrocha (approved):
OK, funny story, Github was showing an outdated version of the code and my comments do not apply. Thank you @cryptoAtwill for syncing, we wouldn't have realized.

dnkolegov (approved):

PR #310, opened 2023-12-08
Add Aderyn scanner

dnkolegov:
This PR adds [Aderyn](https://github.com/Cyfrin/aderyn) scanner and fixes all issues it has found.

https://github.com/consensus-shipyard/ipc-pm/issues/86


src/lib/CrossMsgHelper.sol
```diff
@@ -61,22 +61,22 @@ library CrossMsgHelper {
             });
     }
 
-    function toHash(CrossMsg memory crossMsg) internal pure returns (bytes32) {
+    function toHash(CrossMsg memory crossMsg) external pure returns (bytes32) {
```
adlrocha, line 64: Why do we need to change the visibility of this function? I guess this was a suggestion from the static analyzer, but did it give any reason?

adlrocha (approved):

src/lib/CrossMsgHelper.sol
dnkolegov: This function is not called internally, so the scanner suggests that it should be called externally only.

Issue #311, opened 2023-12-08
Killing and leaving a subnet with confirmation


dnkolegov:
All validators must leave the subnet to kill it.
But to leave the subnet the validators also needed to confirm the next configuration via `submitCheckpoint` call. To do that, we need the majority of votes in a checkpoint.

It works in tests because we use a mock that confirms everything.

It does not work in practice.

adlrocha:
This is a great question, and is definitely a chicken-and-egg problem. The reason why it is like this for now is because it not practically possible to kill a subnet in many cases, so we didn't pay much attention to it. In order to kill the subnet the circulating supply of the subnet needs to be exactly zero which quite a strong requirement.

My proposal right now would be that subnets can't be killed unless we use the `forceKill` function that we are suggesting for the disaster recovery feature. This would mean that if all validators leave a subnet, the subnet would be inactive, but any user that is elegible to be a miner in the subnet is able to join and keep operating it. If the subnet wants to be killed the `owner` would have to `forcekill` it to pause it. 

In the future we will probably need to open a discussion on what is the best thing to do:
- Keep it as-is.
- Leverage the escape hatch functionality and let users with remaining circulating supply vote if the subnets nedes to be killed and the supply returned to its owners.
- Allow the last validator leaving the subnet choose if it wants to close it as soon as he is out.
- etc.

aakoshh:
I agree that it needs research by people smarter than me. 

The idea that a subnet should have 0 circulating supply is a nice one, but it's impossible to enforce: a single user who lost their keys could mean there's no way to release all funds in any legal way. 

Here's where we could say: if the parent subnet also contained all the balances of every user in the child subnet, it could release those when it's killed and then nobody is left out of pocket.

What's not clear to me is how this would work with the [custom ERC20](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/313) feature. Say Alice brings 1 FIL into a subnet `/r0/s1` and turns it into 10 EXC (ExcampleCoin) to use it in `/r0/s1/s2`; she does this by buying EXC from Bob on `/r0/s1` and funding herself in `/r0/s1/s2`. If we wanted to kill `/r0/s1/s2`, then every user in there should have an up-to-date balance in the `/r0/s1/s2` subnet actor living in `/r0/s1` maintained in EXC. And if we wanted to kill `/r0/s1` itself, then what do these users have? They don't hold FIL, so how should they be compensated on `/r0`?

[...]

PR #312, opened 2023-12-08
Remove mocks from integration tests

dnkolegov:
This PR removes mocks from integration tests.


test/SubnetActorDiamond.t.sol
```diff
@@ -511,40 +513,72 @@ contract SubnetActorDiamondTest is Test {
...
+        saManager.leave();
+        vm.stopPrank();
+
+        _confirmChange(validator2, privKey2, validator3, privKey3);
+
+        require(!saGetter.isActiveValidator(validator1), "validator 1 is active");
+        require(saGetter.isActiveValidator(validator2), "validator 2 is not active");
+        require(saGetter.isActiveValidator(validator3), "validator 3 is not active");
+
+        // TODO:
```
adlrocha, line 571: Let's discuss it, but maybe we can remove `kill` for now as discussed here: https://github.com/consensus-shipyard/ipc-solidity-actors/issues/311#issuecomment-1849539344

adlrocha (approved):
Overall LGTM, thanks!

test/SubnetActorDiamond.t.sol
dnkolegov: Already removed

PR #313, opened 2023-12-10
ERC20: enable a subnet's supply to be determined by an ERC20 token in the parent.

raulk:
## Context

Various IPC users seek to create dedicated crypoeconomies backed by their own token. They'd like to launch their token on another network and inject it into their subnet as native circulating supply, so it can be used (a) to pay for gas, and (b) as the native coin for value transfers.

## Contribution

This PR introduces the notion of the "supply source" at the parent subnet. This configuration object determines the kind of supply backing a child subnet. It can be "native" or "ERC20". In the latter case, the token address must be provided. We verify that the address indeed corresponds to an ERC20 token.

We also add a `funcWithToken()` external non-payable function to the `GatewayManagerFacet`. This is analogous to `fund()`, but for token supply (where the amount must be passed as a parameter).

We block message propagation (when the feature flag is enabled) whenever ERC20 networks are involved, as a sender, receiver, or intermediary network.

## Non-goals

This PR does NOT introduce the ability to use an ERC20 token as validator stake. We don't expect that feature to be useful at all, since a major benefit of IPC is that subnets draw economic security from the value of the collateral locked, which is denominated in the root coin.

## Design doc

https://www.notion.so/pl-strflt/ERC20-Native-coin-conversion-eab9267445b6411d96ec17ef02a6510e?pvs=4

## TODO

- [x] Prevent horizontal subnet xnet when subnet is using ERC20 supply.
- [x] Releasing/withdrawing funds.
- [x] ~Unit tests.~ (preferring integration tests with a real ERC20 token involved)
- [x] Integration tests (with TODOs for pending test cases)


src/SubnetActorDiamond.sol
```diff
@@ -51,6 +53,16 @@ contract SubnetActorDiamond {
         if (params.powerScale > 18) {
             revert InvalidPowerScale();
         }
+        if (params.supplyStrategy.kind == SupplyKind.ERC20) {
+            if (params.supplyStrategy.tokenAddress == address(0)) {
+                revert InvalidERC20Address();
+            }
+            // We require that the ERC20 token exists beforehand.
+            // The call to balanceOf will revert if the supplied address does not exist, or if it's not an ERC20 contract.
```
adlrocha, line 61: Nice trick!

src/gateway/GatewayManagerFacet.sol
```diff
@@ -154,6 +165,31 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
...
+        // Transfer the specified token amount to the gateway to lock it.
+        IERC20(supplyStrategy.tokenAddress).safeTransferFrom({from: msg.sender, to: address(this), value: amount});
+
+        // Create the top-down message to mint the supply in the subnet.
+        CrossMsg memory crossMsg = CrossMsgHelper.createFundMsg({
+            subnet: subnetId,
+            signer: msg.sender,
+            to: to,
+            value: amount,
+            fee: 0 // injecting funds into a subnet should is free
```
adlrocha, line 186:
```suggestion
            fee: 0 // injecting funds into a subnet is free
```

src/errors/IPCErrors.sol
```diff
@@ -61,12 +62,13 @@ error PQEmpty();
 error ParentFinalityAlreadyCommitted();
 error PostboxNotExist();
 error SignatureReplay();
+error SubnetAlreadyBootstrapped();
 error SubnetAlreadyKilled();
 error SubnetNotActive();
 error SubnetNotFound();
+error UnexpectedSupplyStrategy();
```
maciejwitowski, line 69: I'd suggest to rename this to `IllegalSupplyStrategy` to be consistent with other `Illegal...` errors related to incorrect function params.

src/gateway/GatewayManagerFacet.sol
```diff
@@ -154,6 +165,31 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
         LibGateway.commitTopDownMsg(crossMsg);
     }
 
+    function fundWithToken(SubnetID calldata subnetId, FvmAddress calldata to, uint256 amount) external nonReentrant {
```
dnkolegov, line 168: Should we validate the address `to` here? Zero address?

src/lib/SupplySourceHelper.sol
```diff
@@ -0,0 +1,101 @@
...
+    function validate(SupplySource memory supplySource) internal view {
+        if (supplySource.kind == SupplyKind.ERC20) {
+            if (supplySource.tokenAddress == address(0)) {
+                revert InvalidERC20Address();
+            }
+            // We require that the ERC20 token contract exists beforehand.
+            // The call to balanceOf will revert if the supplied address does not exist, or if it's not an ERC20 contract.
+            // Ideally we'd use ERC165 to check if the contract implements the ERC20 standard, but the latter does not support supportsInterface().
+            IERC20 token = IERC20(supplySource.tokenAddress);
+            token.balanceOf(address(this));
```
snissn, line 36: I think I understand what you're doing with the balanceOf check but it's a little confusing to use address(this) -- could we use something static like the burn address? I think it just makes it easier to read to make a point that we just want to see if it succeeds for any address, not specifically checking if the  SubnetActorDiamond's call to balanceOf with its own address succeeds. Nit - happy with this being declined if that makes it seem weirder

test/integration/GatewayDiamondToken.t.sol
```diff
@@ -0,0 +1,124 @@
...
+import {GatewayRouterFacet} from "../../src/gateway/GatewayRouterFacet.sol";
+import {SubnetActorManagerFacet} from "../../src/subnet/SubnetActorManagerFacet.sol";
+import {SubnetActorGetterFacet} from "../../src/subnet/SubnetActorGetterFacet.sol";
+import {DiamondLoupeFacet} from "../../src/diamond/DiamondLoupeFacet.sol";
+import {DiamondCutFacet} from "../../src/diamond/DiamondCutFacet.sol";
+import {LibDiamond} from "../../src/lib/LibDiamond.sol";
+
+import {IERC20} from "openzeppelin-contracts/token/ERC20/IERC20.sol";
+
+contract GatewayActorDiamondTest is Test, IntegrationTestBase {
```
snissn, line 48: contract name doesn't match filename

snissn:
For tests you mentioned them being incomplete, here are the tests i can think of to add:

- assert fail funding a native subnet with erc20
- assert fail funding an erc20 subnet with native
- test using a real erc20 as opposed to a mock
- test withdrawing from the subnet (is that possible?) 
- built hardhat entry points and makefile commands that execute code and use that as a test

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -98,27 +100,47 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
...
+        // as soon as we learn that one of the networks involved use an ERC20 supply source.
+        // This will block propagation on the first step, or the last step.
+        //
+        // TODO IPC does not implement fault handling yet, so if the message fails
+        //  to propagate, the user won't be able to reclaim funds. That's one of the
+        //  reasons xnet messages are disabled by default.
+        bool reject;
+        if (isLCA) {
+            // We're connecting both networks, so we check them directly.
+            reject = from.getActor().hasSupplyOfKind(SupplyKind.ERC20);
```
adlrocha, line 116: I think this is wrong, if we assume that multi-level is enabled you may be committing a message locally in a subnet, but you don't have access to the actors of the `from` or the `to` because you are an intermediate subnet and they have nothing to you with you. You only need to forward the message, but you know nothing about the source or origin subnets. It should be fine for now, but we should add a not or something to flag it and fix it in the future.

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -98,27 +100,47 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
...
+            return (shouldBurn = false);
         }
 
-        LibGateway.commitTopDownMsg(crossMessage);
-
-        return (shouldBurn = false);
+        // Else, commit a bottom up message.
+        LibGateway.commitBottomUpMsg(crossMessage);
+        // gas-opt: original check: value > 0
+        return (shouldBurn = crossMessage.message.value != 0);
```
adlrocha, line 143: This assumes that the message value includes the fee to be paid (which is currently the case, but worth noting in case we change the way it works in the future.

src/gateway/GatewayManagerFacet.sol
raulk: I don't think so? `0x00..00` is generally used as a burn address and I don't see why we should specifically prevent deposits straight into the `0x00..00` address. I'm not aware of ERC20 code that prevent such transfers in Ethereum. cc @snissn

src/lib/SupplySourceHelper.sol
```diff
@@ -16,6 +17,13 @@ library SupplySourceHelper {
     error UnexpectedSupplySource();
     error UnknownSupplySource();
 
+    /// @notice Assumes that the address provided belongs to a subnet rooted on this network,
+    ///         and checks if its supply kind matches the provided one.
+    ///         It reverts if the address does not correspond to a subnet actor.
+    function hasSupplyOfKind(address subnet, SupplyKind compare) internal view returns (bool) {
```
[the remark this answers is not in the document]
raulk: Ok

src/lib/SupplySourceHelper.sol
raulk: Sure, I don't see any harm in doing either. I'll set `address(0)` here.

src/lib/SupplySourceHelper.sol
```diff
@@ -0,0 +1,101 @@
...
+        if (supplySource.kind != kind) {
+            revert UnexpectedSupplySource();
+        }
+    }
+
+    /// @notice Locks the specified amount into custody.
+    function lock(SupplySource memory supplySource, address from, uint256 value) internal {
+        if (supplySource.kind == SupplyKind.ERC20) {
+            IERC20 token = IERC20(supplySource.tokenAddress);
+            token.safeTransferFrom({from: from, to: address(this), value: value});
```
adlrocha, line 51: This assumes a previous approve from the user to the gateway, right? Should we document it explicitly somewhere so we don't forget to add it to the docs and is clear for auditors (or anyone else reading the code?)

adlrocha (approved):
Left a few comments. Apart from something that I feel may be wrong when handling multi-level propagations (and that shouldn't be blocking) the implementation of the ERC20 side of things LGTM. I see that at least the happy path and some failure cases are covered in the tests, so should be good to go (at least for the audits)

src/lib/SupplySourceHelper.sol
raulk: Yes, this is also how bridges work more generally, but agree we should document.

src/gateway/GatewayMessengerFacet.sol
raulk: @adlrocha yep, this check only seems to catch sibling to sibling xnet comms (where the first hop is to your LCA). I'll adjust this. Thanks for catching!

raulk: On the bottom-up path or the pivot, I'd love to catch if the `arrivingFrom` network was an ERC20 network and reject the message then. However, unless we save the forwarder network in the postbox, we can't do this under `propagate`. I don't want to do it in `_applyMsg` because that would make a whole batch fail. I'll open an issue for this, as I don't think it's worth refactoring the postbox _right now_ to store _both_ the `arrivingFrom` and the `CrossMsg`.

raulk: https://github.com/consensus-shipyard/ipc-solidity-actors/issues/393

src/gateway/GatewayMessengerFacet.sol
raulk: I don't think I changed anything here. Do you want to open an issue to track this?

scripts/deploy-gateway.template.ts
```diff
@@ -1,8 +1,8 @@
 /* global ethers */
-/* eslint prefer-const: "off" */
 
-import hre, { ethers } from 'hardhat'
+/* eslint prefer-const: "off" */
 import { deployContractWithDeployer, getTransactionFees } from './util'
+import hre, { ethers } from 'hardhat'
```
raulk, line 5: The husky prettier hook insists on making these changes 🤷

src/lib/SupplySourceHelper.sol
raulk: Added soldocs in the `IGateway` interface.

test/integration/GatewayDiamondToken.t.sol
```diff
@@ -0,0 +1,270 @@
...
+        gwRouter.execBottomUpMsgBatch(batch);
+        assertEq(token.balanceOf(recipient), 8);
+    }
+
+    function test_propagation() public {
+        // TODO:
+        // 1. Test that propagation is rejected when sender is ERC20.
+        // 2. Test that propagation is rejected when receiver is ERC20.
+        // 3. Test that propagation is rejected when an intermediary subnet is ERC20.
+    }
```
raulk, line 243: TODO, will open an issue to track these missing tests.

src/gateway/GatewayManagerFacet.sol
snissn: 0x0 has a "special case-ness" in ethereum and i'm not 100% familiar with all of the mechanics to it. Here's some discussion -- https://forum.openzeppelin.com/t/removing-address-0x0-checks-from-openzeppelin-contracts/2222/2

interestingly it seems to regard 0x0 as the "burn address" and wants to separate semantics around transfer and burn. we may want to consider 0x0 as our burn address and welcoming this standard

raulk: The comments in that discussion sum it up pretty well. Various standards have coalesced towards treating the 0x0 address as the creator and burn sink of things. I think blocking 0x0 at this level may hinder legitimate user flows down the line. That said, enforcing the restriction now and then loosening it up is more backwards compatible than the other way around...

raulk:
@snissn 

> * assert fail funding a native subnet with erc20
> * assert fail funding an erc20 subnet with native
> * test using a real erc20 as opposed to a mock
> * test withdrawing from the subnet (is that possible?)

Implemented!

> built hardhat entry points and makefile commands that execute code and use that as a test

Could you elaborate?

raulk:
I'm seeing flaky tests failing at the diamond selector generation step, which is a Python script that calls Forge from within tests via FFI. It's flaky because it appears to fail occasionally, and with a different test each time.

I've raised [a question in Slack](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1702944365013549) to @dnkolegov to understand why we perform selector generation at test time, when we'll anyway need to do this during deployment or sooner.

On surface, it seems like a codegen thing at build time, so curious what other reasons led us to choose this solution.

raulk:
For now, I'll go ahead and merge. `dev` seems to be broken anyway due to failed ABI generation and flaky tests.

src/gateway/GatewayManagerFacet.sol
aakoshh: Maybe worth considering https://github.com/consensus-shipyard/fendermint/pull/409 has introduced 0x0 for the system address, so at least it works as a source. Hopefully the fact that it doesn't have an actual Ethereum account doesn't interfere with using it as a burn address.

PR #319, opened 2023-12-11
IPC-316: Put L2+ support behind a feature flag and remove release() cross-net fee

### LINKED ISSUE #316: M3: Disable L2+ primitives and general purpose cross-net messages
* Disable `sendCross` and `propagate`, and prevent the deployment of L2+ subnets.

adlrocha:
Fixes #316, #317.

This is the first PR for the "Hardening Cross-net Messages" implementation for M3.
* It includes a new `l2PlusSupport` that limits the deployment of L2+ subnets. 
* Removes the minimum fee required for release.
* Disables general-purpose cross-net message operations.

## Question for reviewers.
* I've commented some lines in the tests because they may be useful in the future we decide to re-enable these features (should I remove them?).
* We still have `[minCrosNetFee](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/9c85f9d110ba10737d258094085e70bb9f0ad54d/src/GatewayDiamond.sol#L23)` as a construct parameters for the gateway. We currently not use this value anymore. Should we keep it for future reference or directly remove it (as it may be confusing).

## Next

In the next PR: 
* I will decouple checkpoints from cross-net execution. 
* I will disable relayer rewards to adapt to the current design where there are no fees or incentives for them (at least for now)


src/gateway/GatewayManagerFacet.sol
```diff
@@ -23,6 +23,12 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
     /// @dev The subnet can optionally pass a genesis circulating supply that would be pre-allocated in the
     /// subnet from genesis (without having to wait for the subnet to be spawned to propagate the funds).
     function register(uint256 genesisCircSupply) external payable {
+        // If L2+ support is not enable, only allow the registration of new
+        // subnets in the root
+        if (!s.l2PlusSupport && !s.networkName.isRoot()) {
+            revert MethodNotAllowed();
```
dnkolegov, line 29: I would suggest using another error - L2NotSupported.
MethodNotAllowed targets cross-net messaging

src/gateway/GatewayManagerFacet.sol
```diff
@@ -167,8 +172,8 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
             subnet: s.networkName,
             signer: msg.sender,
             to: to,
-            value: msg.value - fee,
-            fee: fee
+            value: msg.value,
+            fee: 0 // making releases free of fee (at least for now)
```
dnkolegov, line 176: This is the third or fourth time when we have changed that. 
Can we implement different strategies and call them from `release` instead of rewriting the code?
```
release() {
   releaseWithNoFee()
}
```

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -24,7 +24,11 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
      *
      * @param crossMsg - a cross-message to send
      */
-    function sendCrossMessage(CrossMsg calldata crossMsg) external payable validFee(crossMsg.message.fee) {
+    function sendCrossMessage(CrossMsg calldata crossMsg) external payable {
+        if (!s.l2PlusSupport) {
+            revert MethodNotAllowed();
```
dnkolegov, line 29: The same here: let's use the explicit error.

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -46,6 +50,10 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
      * @param msgCid - the cid of the cross-net message
      */
     function propagate(bytes32 msgCid) external payable {
+        if (!s.l2PlusSupport) {
+            revert MethodNotAllowed();
```
dnkolegov, line 54: The same here: let's use the explicit error.

dnkolegov:
> I've commented some lines in the tests because they may be useful in the future we decide to re-enable these features (should I remove them?).

Can we run these tests if l2+ is not enabled? That would be better because that requires us to do something with tests when we remove l2plus settings from the code.

> We still have [minCrosNetFee](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/9c85f9d110ba10737d258094085e70bb9f0ad54d/src/GatewayDiamond.sol#L23) as a construct parameters for the gateway. We currently not use this value anymore. Should we keep it for future reference or directly remove it (as it may be confusing).

The same idea

src/gateway/GatewayManagerFacet.sol
adlrocha: I am using this error for all functionalities that have been disabled as part of M3. When setting `permissioned` I also throw this error if `join`, `leave`, etc. is thrown. I didn't write `MethodNotAllowed` originally to only target cross-net message. If I change this error I should also change how I am using it in that case. Let me know what makes more sense to you.

src/gateway/GatewayManagerFacet.sol
adlrocha: Tell me about it. That being said, I don't see what your proposal gives us. We would end up with a lot of outdated code that we may not touch again.

src/gateway/GatewayManagerFacet.sol
dnkolegov: Ok

dnkolegov (approved):

src/GatewayDiamond.sol
```diff
@@ -51,6 +51,9 @@ contract GatewayDiamond {
         ds.supportedInterfaces[type(IDiamondCut).interfaceId] = true;
         ds.supportedInterfaces[type(IDiamondLoupe).interfaceId] = true;
 
+        // Disable always L2+ cross-net interactions (for now).
+        s.l2PlusSupport = false;
```
raulk, line 55: I would generalise this into `maxTreeDepth` and have it be a number. In that form, it can be a generally useful configuration parameter for subnet operators that, e.g. only want to support 1, 2, N levels under their subnet.

Another perhaps more useful form is to make this a `maxLevelsBelow`, and have the protocol decrement the value in the gateway of a child subnet. It's more functional, but maybe more fragile as it has more moving parts.

src/gateway/GatewayManagerFacet.sol
```diff
@@ -23,6 +23,12 @@ contract GatewayManagerFacet is GatewayActorModifiers, ReentrancyGuard {
     /// @dev The subnet can optionally pass a genesis circulating supply that would be pre-allocated in the
     /// subnet from genesis (without having to wait for the subnet to be spawned to propagate the funds).
     function register(uint256 genesisCircSupply) external payable {
+        // If L2+ support is not enable, only allow the registration of new
```
raulk, line 26:
```suggestion
        // If L2+ support is not enabled, only allow the registration of new
```

src/gateway/GatewayManagerFacet.sol
raulk: I think you can have a `string reason` parameter in the error type to communicate a brief reason.

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -24,7 +24,11 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
      *
      * @param crossMsg - a cross-message to send
      */
-    function sendCrossMessage(CrossMsg calldata crossMsg) external payable validFee(crossMsg.message.fee) {
+    function sendCrossMessage(CrossMsg calldata crossMsg) external payable {
+        if (!s.l2PlusSupport) {
```
raulk, line 28: I think this deserves _another_ feature flag. In the future, we may have multi-level subnets that do not support arbitrary xnet messages.

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -24,7 +24,11 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
      *
      * @param crossMsg - a cross-message to send
      */
-    function sendCrossMessage(CrossMsg calldata crossMsg) external payable validFee(crossMsg.message.fee) {
+    function sendCrossMessage(CrossMsg calldata crossMsg) external payable {
```
raulk, line 27: The name of this method is ambiguous.

- We use the struct `CrossMsg` to convey even fund and withdrawal messages.
- We don't want to disable those! We want to disable only arbitrary messages.
- But the name of this method and the condition makes it sound like we're disabling _every_ cross message including fund and withdrawal system messages.

I think we should distinguish between user messages and system messages.

A broader nit is the term "cross message" -- it's missing an object: what is the message crossing exactly? I think it should be "cross-net message". I'd abbreviate cross-net as "xnet": `sendXnetMessage`.

Propose to call this: `sendUserXnetMessage`.

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -46,6 +50,10 @@ contract GatewayMessengerFacet is GatewayActorModifiers {
      * @param msgCid - the cid of the cross-net message
      */
     function propagate(bytes32 msgCid) external payable {
+        if (!s.l2PlusSupport) {
```
raulk, line 53: Same, use another feature flag. It may even be a dedicated propagation feature flag?

src/GatewayDiamond.sol
adlrocha: I don't think I agree here. Subnets are user-defined, so they can only make local decisions about their own configuration. Thus, a subnet may choose to enable or disable child subnets anchoring to it, but at this point I don't think we can reliably enforce a `maxDepth` from them (at least if we want to keep subnet deployment decentralized).

That being said, I agree that the name for the feature flag may not be the best, is probably better to call it `allowChildSubnets` or something like that. I will update the code accordingly.

adlrocha: Good point, but I just want to note that while we could suggest child subnets the implementation that they deploy for the gateway in their network, we can't prevent them from changing the code to disable this `maxDepth` (as long as we want the deployment of subnets to be decentralized). Sunets can only enforce local decisions and configurations on their own network.

In any case, I will include this flag  as it may be useful in the future, but I am afraid it may have a limited use (I hope I am wrong).

src/gateway/GatewayMessengerFacet.sol
adlrocha: No strong opinion here. This method hasn't been used extensively. Fixing accordingly

src/gateway/GatewayMessengerFacet.sol
adlrocha: fixed

src/gateway/GatewayMessengerFacet.sol
adlrocha: fixed

raulk (commented):
For extra safety, should we update the gateway router to prevent messages so that we only allow execution of messages whose to == us, i.e. prohibiting forwarding for now?

adlrocha:
> For extra safety, should we update the gateway router to prevent messages so that we only allow execution of messages whose to == us, i.e. prohibiting forwarding for now?

In M2 we had this, where actually the API for `fund` and `release` didn't include an address as an argument, so you could only send funds to yourself, but we ended up enabling it because some users mentioned that it was a PITA having to send two transactions to move funds from a subnet to an address they don't own. It may reduce minimally the attack surface, but I don't know if it is worth the pain for users (considering the already limiting capabilities of M3).

That being said, happy to be convinced otherwise :)

src/gateway/GatewayMessengerFacet.sol
```diff
@@ -6,25 +6,32 @@ import {BURNT_FUNDS_ACTOR} from "../constants/Constants.sol";
...
 import {SubnetID} from "../structs/Subnet.sol";
-import {InvalidCrossMsgFromSubnet, InvalidCrossMsgDstSubnet, CannotSendCrossMsgToItself, InvalidCrossMsgValue} from "../errors/IPCErrors.sol";
+import {InvalidCrossMsgFromSubnet, InvalidCrossMsgDstSubnet, CannotSendCrossMsgToItself, InvalidCrossMsgValue, MethodNotAllowed} from "../errors/IPCErrors.sol";
 import {SubnetIDHelper} from "../lib/SubnetIDHelper.sol";
 import {LibGateway} from "../lib/LibGateway.sol";
 import {StorableMsgHelper} from "../lib/StorableMsgHelper.sol";
 import {FilAddress} from "fevmate/utils/FilAddress.sol";
 
+string constant ERR_GENERAL_CROSS_MSG_DISABLED = "Support for general-purpose cross-net messages is disabled";
+string constant ERR_MULTILEVEL_CROSS_MSG_DISABLED = "Support for general-purpose cross-net messages is disabled";
```
raulk, line 16: Was the error text supposed to be different here?

src/gateway/GatewayMessengerFacet.sol
raulk, line 1: Just wanted to confirm: `fund` and `release` messages do not currently support multi-level anyway, right? i.e. I cannot fund an account/contract in subnet 3 from subnet 1 if via path `/1/2/3`?

raulk (approved):

src/gateway/GatewayMessengerFacet.sol
adlrocha: They were supposed to, good catch. Fixed in the next commit.

src/gateway/GatewayMessengerFacet.sol
adlrocha: Correct! They are single-level by definition.

PR #323, opened 2023-12-12
Pausable

cryptoAtwill:
Add `pausable` to subnet actor such that operations with `whenNotPaused` will be blocked when necessary.


src/lib/LibPausable.sol
```diff
@@ -0,0 +1,42 @@
...
+    error AlreadyPaused();
+
+    modifier whenNotPaused() {
+        PausableStorage storage s = pausableStorage();
+        if (s.paused) revert AlreadyPaused();
+        _;
+    }
+
+    /// @notice sets if to pause the contract
+    function isPausabled() external view returns(bool) {
```
dnkolegov, line 22: I think it should be named `isPaused`

dnkolegov: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Pausable.sol

src/lib/LibPausable.sol
```diff
@@ -0,0 +1,42 @@
...
+        PausableStorage storage s = pausableStorage();
+        return s.paused;
+    }
+
+    /// @notice sets if to pause the contract
+    function setPausable(bool paused) internal {
```
dnkolegov, line 28: Why do you need an argument? Why do we need setPausable(false)?
It should be named like `pause()`.
I suggest using functions names and semantics from https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Pausable.sol

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -238,7 +252,7 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     /// @notice method that allows a validator to leave the subnet
     /// @dev it also return the validators initial balance if the
     /// subnet was not yet bootstrapped.
-    function leave() external nonReentrant notKilled {
+    function leave() external whenNotPaused nonReentrant notKilled {
```
dnkolegov, line 255: `nonreentrant` always should be first one

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -286,17 +300,17 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
     }
 
     /// @notice Validator claims their released collateral
-    function claim() external nonReentrant {
+    function claim() external whenNotPaused nonReentrant {
```
dnkolegov, line 303: nonreentrant always should be the first one

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -286,17 +300,17 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Reentran
...
         LibStaking.claimCollateral(msg.sender);
     }
 
     /// @notice Relayer claims its reward
-    function claimRewardForRelayer() external nonReentrant {
+    function claimRewardForRelayer() external whenNotPaused nonReentrant {
```
dnkolegov, line 308: nonreentrant always should be the first one

dnkolegov (changes requested):
I suggest using https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Pausable.sol as a reference: use functions names and semantics.

dnkolegov:
> Add `pausable` to subnet actor such that operations with `whenNotPaused` will be blocked when necessary.

Are we going to add pausability for Gateways?

cryptoAtwill:
close temporarily to avoid generating `out` and `binding`.

cryptoAtwill:
@dnkolegov I have updated the code based on your feedbacks. Please take another look.

dnkolegov (approved):

PR #326, opened 2023-12-12
IPC-318: Decouple checkpoint from messages

adlrocha:
Builds upon: #319 
Part 1 of impl. for https://github.com/consensus-shipyard/ipc-solidity-actors/issues/318

This PR is the second part of "Hardening Cross-net messages", it was getting too long so I decided to split it into several PRs. This one:
- Decouples messages from checkpoint submission.
- Abstract the logic to submit signatures and create quorum over objects so it can be re-used by bottom-up message batches in the next PR.
- Fixes a bug where epochs and height were being typed as `uint64` when in Solidity it is a `uint256` (I feel this was dragged from the translation of the Rust actors into Solidity, as `ChainEpoch` in Filecoin is represented through a `uint64`). 

## Next
- Implement the creation and submission of `QuorumBundles` for the execution of bottom-up messages. 

Note that this change may include several breaking changes to the deployment scripts, Fendermint and IPC (but we should probably address those once the monorepo is ready to make our lives easier).


.github/workflows/prettier.yml
```diff
@@ -20,21 +19,20 @@ jobs:
...
-#        with:
-#          commit_message: Apply Prettier formatting
-#          branch: ${{ github.head_ref }}
+          node-version: '14'
+      - name: Install Prettier and Prettier Solidity Plugin
+        run: npm install --save-dev prettier prettier-plugin-solidity
+      - name: Run Prettier on JavaScript/TypeScript/JSON/CSS/MD
+        run: npx prettier --write '**/*.{js,jsx,ts,tsx,json,css,md}'
+      - name: Run Prettier on Solidity Files
+        run: npx prettier --write '**/*.sol'
```
aakoshh, line 31: Doesn't `make fmt` do this now?

raulk: I think this PR pulled in stale configs during merge (see my comment above)

PR #328, opened 2023-12-12
Solidity version 8.23

snissn:
upgrade solidity for bugfix security reasons


snissn:
closing for now as hardhat claims only 0.8.22 is the highest support version. please see https://github.com/consensus-shipyard/ipc-solidity-actors/pull/334 for the newest version

PR #330, opened 2023-12-13
IPC-318: Submit bottomup batches for execution

adlrocha:
Builds upon #326 
Part 2 of impl. for https://github.com/consensus-shipyard/ipc-solidity-actors/issues/318

## Implementation
This PR decouples the execution of bottom-up messages form the submission of checkpoints.
* When new bottom-up messages are committed they are included into a `BottomUpMsgBatch`. 
* Every `bottomUpMsgBatchPeriod`, validators can create a new batch to run a quorum over it and submit it to the parent. 
* If the batch reaches a `maxMsgsPerBatch` a `NewBatch` event is emitted notifying validators that they should run a quorum over a batch and submit it to the parent before the next checkpoint period. 
* The submission of message batches for execution is performed in the same way it is done for checkpoints. 
  * Validators run a quorum over the batch. 
  * When the quorum is reached a new `QuorumEvent` is emitted to notify relayers. 
  * Relayers pick up the batch and submit it to the parent. 
  
## Changes in Fendermint
This PR entails several changes in the API with the contracts and in the way Fendermint should behave.
* We should adapt the new checkpoint format. 
* Include a new process that listens to events for `NewBatch` or that creates a new batch every batch period.
* This process should add a signature for the validator in the batches were we are currently running a quorum over.


src/GatewayDiamond.sol
```diff
@@ -65,6 +66,13 @@ contract GatewayDiamond {
         s.minCrossMsgFee = params.msgFee;
         s.majorityPercentage = params.majorityPercentage;
         s.checkpointQuorumMap.retentionHeight = 1;
+        s.bottomUpMsgBatchQuorumMap.retentionHeight = 1;
+
+        // BottomUpMsgsBatch config parameters.
+        // NOTE: Let's fix them for now, but we could make them configurable
+        // through the gateay constructor in the future.
```
maciejwitowski, line 73:
```suggestion
        // through the gateway constructor in the future.
```

src/SubnetActorDiamond.sol
```diff
@@ -73,6 +74,12 @@ contract SubnetActorDiamond {
         s.currentSubnetHash = s.parentId.createSubnetId(address(this)).toHash();
         s.permissioned = params.permissioned;
 
+        // BottomUpMsgsBatch config parameters.
+        // NOTE: Let's fix them for now, but we could make them configurable
+        // through the gateay constructor in the future.
```
maciejwitowski, line 79:
```suggestion
        // through the gateway constructor in the future.
```

src/gateway/GatewayGetterFacet.sol
```diff
@@ -167,19 +175,39 @@ contract GatewayGetterFacet {
         return checkpoints;
     }
 
+    /// @notice get the incomplete checkpoints
```
maciejwitowski, line 178:
```suggestion
    /// @notice get the incomplete batches of bottom-up messages
```

src/lib/LibGateway.sol
```diff
@@ -69,6 +91,26 @@ library LibGateway {
...
     }
 
+    /// @notice stores bottom-up batch
+    function storeBottomUpMsgBatch(
+        BottomUpMsgBatch memory batch
+    ) internal {
+        GatewayActorStorage storage s = LibGatewayActorStorage.appStorage();
+        BottomUpMsgBatch storage b = s.bottomUpMsgBatches[batch.blockHeight];
+        b.subnetID = batch.subnetID;
+        b.blockHeight = batch.blockHeight;
```
maciejwitowski, line 101: Hm, 2 lines above we take `BottomUpMsgBatch ` from the `bottomUpMsgBatches` by `batch.blockHeight`. 
Won't this batch already have `b.blockHeight` equal to `batch.blockHeight`?

dnkolegov (approved):

src/lib/LibGateway.sol
dnkolegov: `b.blockHeight` will be 0, we do. not take batch from storage we take a pointer to storage to be able to store the input batch

PR #332, opened 2023-12-13
CI: do not commit, only report on dirty gen.

raulk:
Fixes #299.

### DIFF
--- a/.github/workflows/bindings.yml
+++ b/.github/workflows/bindings.yml
@@ -29,15 +29,20 @@ jobs:
       - name: Create binding
         run: make compile-abi && make rust-binding
 
-      - name: Git Config
+      - name: Check diff clean
         run: |
-          git config --global user.name 'auto-commit'
-          git config --global user.email 'auto-commit@users.noreply.github.com'
+          git --no-pager diff
+          git --no-pager diff --quiet
 
-      - name: Commit bindings
-        run: |
-          make commit-rust-binding
-
-      - name: Commit ABI
-        run: |
-          make commit-abi
+#      - name: Git Config
+#        run: |
+#          git config --global user.name 'auto-commit'
+#          git config --global user.email 'auto-commit@users.noreply.github.com'
+#
+#      - name: Commit bindings
+#        run: |
+#          make commit-rust-binding
+#
+#      - name: Commit ABI
+#        run: |
+#          make commit-abi
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -31,8 +31,12 @@ jobs:
         run: npx prettier --write '**/*.sol'
       - name: Debugging - Show changed files
         run: git restore package-lock.json && git status && git diff --name-only
-      - name: Commit changes
-        uses: stefanzweifel/git-auto-commit-action@v5
-        with:
-          commit_message: Apply Prettier formatting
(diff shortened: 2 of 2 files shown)


.github/workflows/prettier.yml
```diff
@@ -31,8 +31,12 @@ jobs:
...
       - name: Debugging - Show changed files
         run: git restore package-lock.json && git status && git diff --name-only
-      - name: Commit changes
-        uses: stefanzweifel/git-auto-commit-action@v5
-        with:
-          commit_message: Apply Prettier formatting
-          branch: ${{ github.head_ref }}
+      - name: Check diff clean
+        run: |
+          git --no-pager diff --quiet
```
aakoshh, line 36: Just checking: does the `Makefile` have a target that applies the `prettier` formatting locally?

.github/workflows/bindings.yml
```diff
@@ -29,15 +29,20 @@ jobs:
...
-
-      - name: Commit bindings
-        run: |
-          make commit-rust-binding
-
-      - name: Commit ABI
-        run: |
-          make commit-abi
+          git --no-pager diff
+          git --no-pager diff --quiet
```
aakoshh, line 35: So, I'm not sure why since `bindings` has a `Cargo.lock` file, which perhaps doesn't apply to `build.rs`, but @cryptoAtwill said that this diff was showing differences even when he just generated the bindings locally. The only reason I can think of is if the ethers library was different, so maybe fixing it to a certain version [here](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/9c85f9d110ba10737d258094085e70bb9f0ad54d/binding/Cargo.toml#L10) could help with that.

aakoshh: Or maybe `--locked` has to be added [here](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/9c85f9d110ba10737d258094085e70bb9f0ad54d/Makefile#L14C10-L14C96)

.github/workflows/prettier.yml
aakoshh: Ah [there is](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/9c85f9d110ba10737d258094085e70bb9f0ad54d/Makefile#L34-L35), so maybe the CI build should call `make fmt` instead of repeating a potentially different command.

aakoshh (approved):
Looks fine to me, with the potential caveats I commented about.

.github/workflows/bindings.yml
raulk: Yeah, this may also have to do with code formatting / prettier introduced recently. Not sure what the sequence of local actions @cryptoAtwill did, but these files do contain sourcemaps so they're sensitive to source code formatting.

raulk: I would be surprised if `build.rs` doesn't use the `Cargo.lock` file, but that's another hyptohesis worth exploring. For now, I'll add the `fmt` target as a precondidtion to the abigen Makefile targets.

raulk: Ah, I just saw your comment on --locked. I'll try that! It always puzzles me why this is not the default behaviour.

.github/workflows/prettier.yml
raulk: Just saw your comment ;-)

PR #339, opened 2023-12-14
improvements in xnet message application code paths

raulk:
Changes are atomized. See commit log for details.

### DIFF
--- a/src/gateway/GatewayRouterFacet.sol
+++ b/src/gateway/GatewayRouterFacet.sol
@@ -146,9 +146,9 @@ contract GatewayRouterFacet is GatewayActorModifiers {
     }
 
     /// @notice executes a cross message if its destination is the current network, otherwise adds it to the postbox to be propagated further
-    /// @param forwarder - the subnet that handles the cross message
+    /// @param arrivingFrom - the immediate subnet from which this message is arriving
     /// @param crossMsg - the cross message to be executed
-    function _applyMsg(SubnetID memory forwarder, CrossMsg memory crossMsg) internal {
+    function _applyMsg(SubnetID memory arrivingFrom, CrossMsg memory crossMsg) internal {
         if (crossMsg.message.to.subnetId.isEmpty()) {
             revert InvalidCrossMsgDstSubnet();
         }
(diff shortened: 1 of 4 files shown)


dnkolegov (approved):

adlrocha (approved):
LGTM, thanks!

raulk:
@dnkolegov I'm seeing these test failures:

```
[FAIL. Reason: NoParentForSubnet()] testGatewayDiamond_CommitParentFinality_BigNumberOfMessages() (gas: 112181037)
[FAIL. Reason: NoParentForSubnet()] testGatewayDiamond_Propagate_Works_NoFeeReminder() (gas: 318403)
[FAIL. Reason: NoParentForSubnet()] testGatewayDiamond_Propagate_Works_WithFeeRemainder() (gas: 313152)
```

I suspect this is because these tests are sending top-down messages on the gateway of a root network, which can never happen in reality. Could you take a look and adjust the tests if necessary?

src/gateway/GatewayRouterFacet.sol
```diff
@@ -146,9 +146,9 @@ contract GatewayRouterFacet is GatewayActorModifiers {
     }
 
     /// @notice executes a cross message if its destination is the current network, otherwise adds it to the postbox to be propagated further
-    /// @param forwarder - the subnet that handles the cross message
+    /// @param arrivingFrom - the immediate subnet from which this message is arriving
```
maciejwitowski, line 151: The commit message says "It may not be" instead of "It may be" which IIUC is incorrect and may be confusing in the future.

src/gateway/GatewayRouterFacet.sol
```diff
@@ -188,12 +188,12 @@ contract GatewayRouterFacet is GatewayActorModifiers {
 
     /// @notice applies a cross-net messages coming from some other subnet.
     /// The forwarder argument determines the previous subnet that submitted the checkpoint triggering the cross-net message execution.
```
maciejwitowski, line 192: This line should now say "The arrivingFrom argument ..."

src/gateway/GatewayRouterFacet.sol
raulk: The commit message is logically correct, but I'll reword to make it clearer.

src/gateway/GatewayRouterFacet.sol
raulk: Thanks!

PR #345, opened 2023-12-14
CI: Generate bindings *after* the merge and separate from `out`

aakoshh:
This PR tries to help with the problems we face if the ABI artifacts cannot be built deterministically, leading to unnecessary merge conflicts, by:
* adding `out` to `.gitignore` (and removing previously checked in sources)
* changing the `bindings.yaml` workflow to run after the PR has been merged, so hopefully it never encounters merge conflicts
* generates ABI artifacts into an `.out` directory which we should not touch locally (we work with `out`)
* generate Rust `bindings`, which we should again not touch locally to avoid merge conflicts
* commits and pushes those from CI

CAVEAT: We have to keep an eye on the `dev` build passing and the new commit appearing.

This should keep `bindings` available for the dependant `ipc` and `fendermint` repositories. In the near future when we switch to the monorepo this will be done differently, so this PR might be the quickest way to keep things chugging along.


Makefile
```diff
@@ -2,7 +2,7 @@
 # Deployment
 
 NETWORK ?= auto
-OUTPUT ?= ./out
+OUTPUT ?= .abi
```
raulk, line 5: These aren't just the abis, but the full compiler outputs, incl bytecode and debug information. So `.out` would be a more accurate name.

.gitignore
```diff
@@ -9,6 +9,7 @@ broadcast/
 node_modules
 
 artifacts
+out
```
raulk, line 12: What would this entry be for now?

raulk (approved):

Makefile
aakoshh: Yeah; I only named it like this because we say `compile-abi`, but we can call it `.out`.

.gitignore
aakoshh: What do you mean?

raulk: We no longer have an `out` dir. I guess this is to ignore outputs if the user runs `foundry build` locally (which I think defaults to out/)?

PR #349, opened 2023-12-14
CI: No bindings on push

aakoshh:
Not running the `bindings.yaml` workflow on `push` because it triggers itself recursively.


raulk (approved):

raulk (changes requested):
I think it's better to exclude commits authored by the bot.

adlrocha (approved):

raulk (approved):
I like it!

PR #352, opened 2023-12-14
Use constant for public key size

dnkolegov:
Just use constant for key size


src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -19,6 +19,7 @@ import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.
 import {Address} from "openzeppelin-contracts/utils/Address.sol";
 
 string constant ERR_PERMISSIONED_AND_BOOTSTRAPPED = "Method not allowed if permissioned is enabled and subnet bootstrapped";
+uint256 constant PUBLIC_KEY_LENGTH = 65;
```
raulk, line 22: Let's specify the kind of public key this is (secp256k1?) in the constant name and also add some docs?

raulk (approved):

PR #360, opened 2023-12-15
fix for getterFacet to have the lib LibQuorum

snissn:
running make was failing and this fixes it

i think we should create a make based github action test to catch this


snissn:
not sure what i ran into trouble with package-lock.json because my first commit didn't change anything like that, but this fix may be needed overall anyway due to some unforeseen changes

snissn:
closing for https://github.com/consensus-shipyard/ipc-solidity-actors/pull/371 which does not have unrelated npm changes

PR #367, opened 2023-12-16
  gh workflow check to deploy codebase on local ganache chain

snissn:
opening draft pr to test gh workflow


snissn:
This should be working now except that we first need to merge https://github.com/consensus-shipyard/ipc-solidity-actors/pull/360 so that the build tests actually pass, as a commit to dev was introduced recently causing this to fail

adlrocha (approved):
This is super useful, thank you!

snissn:
This is somewhat blocked on https://github.com/consensus-shipyard/ipc-monorepo/issues/9# since 405 is causing our tests to fail, and it feels incorrect to merge a new test into our CI while it's currently failing due to a different known issue

raulk (approved):
We can merge this here, but we'll need to port the build job to ipc-monorepo shortly!

PR #374, opened 2023-12-17
Mikers/subnet nonce2

snissn:
see https://github.com/consensus-shipyard/ipc-solidity-actors/issues/372


adlrocha (approved):
This is a great point and LGTM. Now I am wondering if we should propagate this change also to other nonces throughout the system like cross-net message nonces, and applied cross-net message nonces.

raulk:
@snissn Hey, in the future can we please make sure to update PR titles so that they're more explanatory of what they're achieving? Note that PR titles and descriptions also make it into the commit log on merge, and "mikers/subnet nonce2" is a commit message that will age very poorly.

PR #387, opened 2023-12-18
Adding tests for static validation

cryptoAtwill:
Adding tests for static validation.

### DIFF
--- a/test/integration/SubnetActorDiamond.t.sol
+++ b/test/integration/SubnetActorDiamond.t.sol
@@ -1246,6 +1246,48 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
         // console.log("callback called");
     }
 
+    function testSubnetActorDiamond_StaticValidation_cannotJoin() public {
+        gatewayAddress = address(gatewayDiamond);
+
+        createSubnetActor(
+            gatewayAddress,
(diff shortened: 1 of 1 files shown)


test/integration/SubnetActorDiamond.t.sol
```diff
@@ -1246,6 +1246,48 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
...
+        (address validator2, bytes memory publicKey2) = TestUtils.deriveValidatorAddress(100);
+        vm.deal(validator2, DEFAULT_MIN_VALIDATOR_STAKE * 2);
+        vm.startPrank(validator2);
+        saManager.join{value: DEFAULT_MIN_VALIDATOR_STAKE / 2}(publicKey2);
+        vm.stopPrank();
+
+        require(saGetter.isActiveValidator(validator1), "not active validator 1");
+        require(saGetter.isActiveValidator(validator2), "not active validator 2");
+
+        // cannot join after bootstrap
```
adlrocha, line 1277: Let's maybe add also a case where we check that we can also not set `setFederatedPower` here. Thanks!

cryptoAtwill: Haha, yeah, there is one, it's at the bottom.

PR #388, opened 2023-12-18
Reduce the integration test base contract state by adding test actors

dnkolegov:


adlrocha (approved):
LGTM. Do you mind rebasing and merging? (let me know if you are not around and I could try to do it myself, but I don't want to mess up anything).

PR #392, opened 2023-12-18
Update registration policy

dnkolegov:
This PR addresses that [discussion](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1701770112802639?thread_ts=1701734407.601389&cid=C04JR5R1UL8) and tries to make `join` simpler for reasoning.

The new implemented policy is as follows:

- a validator must  register itself via `join` by providing a public key
- a validator can't call `join` if it has already been registered
- a validator can't call `stake` if it has not been registered yet
- a validator can't call `join` or `stake` if `msg.value`  is 0


src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -17,6 +17,7 @@ import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.
 import {Address} from "openzeppelin-contracts/utils/Address.sol";
 
 string constant ERR_PERMISSIONED_AND_BOOTSTRAPPED = "Method not allowed if permissioned is enabled and subnet bootstrapped";
+string constant ERR_BOOTSTRAPPED_AND_JOINED = "Method not allowed if subnet bootstrapped and validator has already joined";
```
adlrocha, line 20: I think you want to make this check always, even if the network has already bootstrapped, right? You directly don't want an address to join twice. 
```suggestion
string constant ERR_BOOTSTRAPPED_AND_JOINED = "Method not allowed if validator has already joined";
```

src/lib/LibStaking.sol
```diff
@@ -403,6 +403,14 @@ library LibStaking {
         return s.validatorSet.waitingValidators.contains(validator);
     }
 
+    /// @notice Checks if the provided address is a validator (active or waiting) based on its total collateral.
+    /// @param addr The address to check for validator status.
+    /// @return A boolean indicating whether the address is a validator.
+    function isValidator(address addr) internal view returns (bool) {
```
adlrocha, line 409: This function sounds familiar, do you mind double-checking that it hasn't been implemented somewhere else?

adlrocha (commented):
I think there may be a mismatch between the PR description and what has been actually implemented. 

We don't want a validator to be able to join twice, period, independently of if the subnet is bootstrapped or not, right? I think the implementation is correct but the description is not, but I want to double-check before merging. Thanks!

src/subnet/SubnetActorManagerFacet.sol
dnkolegov: But the method should be allowed for calling several times before the subnet is bootstrapped, right? Because a validator may stake MIN_COLLATERAL amount in several payment.

src/lib/LibStaking.sol
dnkolegov: We have `isActivaValidator` and `isWaitingValidator`

dnkolegov:
> We don't want a validator to be able to join twice, period, independently of if the subnet is bootstrapped or not, right?

If think we have to allow calling `join` several times because a validator may send the collateral in several paements.

src/subnet/SubnetActorManagerFacet.sol
adlrocha: If I am not wrong, `stake` allows you to stake more collateral when the subnet hasn't been yet bootstrapped. The behavior should be the same in both cases, bootstrapped and not bootstrapped, but the implementation flow should change slightly (as we don't need a confirmation to activate the collateral)

adlrocha:
I still think we shouldn't allow joining twice as mentioned in this thread: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/392#discussion_r1430525205. Happy to follow-up sync in case it is easier to discuss.

src/subnet/SubnetActorManagerFacet.sol
```diff
@@ -268,48 +291,37 @@ contract SubnetActorManagerFacet is ISubnetActor, SubnetActorModifiers, Pausable
...
-            // Join only if the sender's address is not already linked to a validator that has joined.
+        } else {
+            // if the subnet has been bootstrapped, join with postponed confirmation.
             LibStaking.setValidatorMetadata(msg.sender, publicKey);
             LibStaking.deposit(msg.sender, msg.value);
-        } else {
-            revert MethodNotAllowed(ERR_BOOTSTRAPPED_AND_JOINED);
         }
+
+        _bootstrapSubnetIfNeeded();
```
adlrocha, line 300: I think this should go between line 293 and 294, we should only do this if it is not bootstrapped. Once bootstrapped `register` shouldn't be called.

test/integration/SubnetActorDiamond.t.sol
```diff
@@ -446,7 +446,7 @@ contract SubnetActorDiamondTest is Test, IntegrationTestBase {
         saManager.stake();
 
         vm.prank(validator);
-        vm.expectRevert(NotStakedBefore.selector);
+        vm.expectRevert((abi.encodeWithSelector(MethodNotAllowed.selector, ERR_VALIDATOR_NOT_JOINED)));
```
adlrocha, line 449: Should we add a test that checks that a validator can't join twice?

adlrocha (approved):
LGTM, thanks!

PR #400, opened 2023-12-19
Remove ERC165 from diamonds constructor

snissn:
remove updates to erc165 related data structure supportedInterfaces because the way we use diamondCut to upgrade our diamond we do not follow erc165


dnkolegov:
> remove updates to erc165 related data structure supportedInterfaces because the way we use diamond-cut to upgrade our diamond we do not follow erc165

Maybe instead of removing the partially implemented interface, we should extend the implementation and support that interface ?

raulk:
Yeah, agree, I think it's best to support ERC-165 as @dnkolegov says, unless I'm missing something. Tracking here: https://github.com/consensus-shipyard/ipc-monorepo/issues/498

PR #401, opened 2023-12-19
remove status and minStake

cryptoAtwill:
This PR removes the minStake and status from the GATEWAY contracts. 

Previously we had this as we only have collateral based PermissionMode and the idea is to use these parameters to control in the Gateway to enforce some control over different subnets. But now we are moving towards different `PermissionMode` and these parameters no longer required.

Note that `minStake` is a parameter applied to all subnets created, but for `PermissionMode.Federated`, there is no collateral requirements, so a `minStake` in the Gateway must be 0. Anything above zero will prevent `Federated` subnet to be register in the gateway. If `minStake` is always 0, just remove this parameter. 

At the same time, `status` will be deprecated as well as collateral alone does not manage the status of the subnet. Overall, subnets are managing their own stake requirements and status, so that the gateway does not really need to be aware of subnet’s status and status is not used anywhere except for listing subnets. Subnet’s status is managed in SubnetActor now, in bootstrapped and killed parameters, Gateway no longer needs to maintain the status.


raulk:
@cryptoAtwill could you elaborate on the context and the motivation, for other team mates and contributors to follow?

raulk:
Thanks for the description, @cryptoAtwill! This is the operative point for me:

> Overall, subnets are managing their own stake requirements and status, so that the gateway does not really need to be aware of subnet’s status and status is not used anywhere except for listing subnets.

raulk (approved):

raulk:
LGTM -- feel free to merge at your convenience, @cryptoAtwill!

dnkolegov (approved):

PR #407, opened 2023-12-20
 precompile diamond function selectors at compile time

snissn:
generating the function selectors is currently failing in our tests in a flaky way. this introduces a python script to compile all of the function selectors into a library which our code can import for our tests


Makefile
```diff
@@ -102,6 +102,10 @@ make coverage-for-mac: | forge
 
 prepare: fmt lint test slither
 
+
+build-selector-library: | forge
+	python scripts/python/build_selector_library.py
```
raulk, line 107: Let's run this in the `prepare` target, which seems to be the main aggregate target here?

Makefile
```diff
@@ -102,6 +102,10 @@ make coverage-for-mac: | forge
 
 prepare: fmt lint test slither
 
+
+build-selector-library: | forge
```
raulk, line 106: We will need to add a CI check to fail when the user has not codegen'ed the selectors.

scripts/python/build_selector_library.py
```diff
@@ -0,0 +1,94 @@
...
+    # Define the file path
+    file_path = 'test/helpers/SelectorLibrary.sol'
+
+    # Write the content to the file
+    with open(file_path, 'w') as file:
+        file.write(selector_storage_content)
+
+
+def generate_solidity_function(contract_selectors):
+    solidity_code = "pragma solidity ^0.8.19;\n"
```
raulk, line 19: Add license?

test/helpers/SelectorLibrary.sol
```diff
@@ -0,0 +1,138 @@
+library SelectorLibrary {
+    function generateSelectors(string memory facetName) public pure returns (bytes4[] memory facetSelectors) {
+        if (keccak256(abi.encodePacked(facetName)) == keccak256(abi.encodePacked("GatewayDiamond"))) {
+            return
+                abi.decode(
+                    hex"00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000",
+                    (bytes4[])
+                );
+        }
```
raulk, line 9: This approach is rather inefficient. I think we'll want a more gas-efficient approach in case we need to ever run this on chain. How about we construct a map in state, populate it in the constructor, and then simply look up inside the map to resolve selectors?

scripts/python/build_selector_library.py
```diff
@@ -0,0 +1,94 @@
...
+    solidity_code = "library SelectorLibrary {\n"
+    solidity_code += "    function generateSelectors(string memory facetName) public pure returns (bytes4[] memory facetSelectors) {\n"
```
raulk, line 21: Would call this function `resolveSelectors`, since we aren't generating the selectors when this function is called any longer.

scripts/python/build_selector_library.py
```diff
@@ -0,0 +1,94 @@
...
+                selectors = get_selectors(filepath + ':' + contract_name)
+                if selectors:
+                    contract_selectors[contract_name] = selectors
+            except Exception as oops:
+                print(f"Error processing {filepath}: {oops}")
+
+
+    # Print the final JSON
+    solidity_library_code = generate_solidity_function(contract_selectors)
+    writeToFile(solidity_library_code)
```
raulk, line 90: I'd rather remove dynamic discovery, and have an explicit list of paths. Less magic and more explicitness for smart contracts is good.

test/helpers/SelectorLibrary.sol
raulk: Alternatively we could generate a function for each facet that returns the hardcoded selectors? Consumer code anyway needs to know the magic strings, which are not introspectable. Functions would give us compilation-time errors. And given that we can pass them as arguments in Solidity, we can also do higher order stuff with that approach?

raulk (changes requested):

raulk:
Interested in @dnkolegov's review here.

dnkolegov (approved):
I have nothing to add to Raul's comments.

Makefile
snissn: One thing worth discussing is how often this needs to be run. Function selectors here should change relatively infrequently and it would be by a developer who is actively changing things. I think that in this case it makes sense to run the script and commit the new file when things change.. I know we have whiplash from our `out` folder auto generating things, but i think simple is good, especially for now, to land this PR then we can change later if it becomes annoying

Makefile
snissn: > We will need to add a CI check to fail when the user has not codegen'ed the selectors.
I think we should assume for now that we have a valid code gen selectors commited to the repo

scripts/python/build_selector_library.py
snissn: 213b5f27 done!

test/helpers/SelectorLibrary.sol
snissn: This code ONLY runs within our tests. I had some designs that initially used a map but that wasn't compatible with solidity libraries and it just was simpler to implement this code as a library because it meant not having to deploy a contract with state. Considering this only runs as a test, the cascading IFs feels a little off, but OKAY. If we start to find a need to have this code on-chain, definitely we would re-implement as a map. I would like to hold off on that until there is a reason to

scripts/python/build_selector_library.py
snissn: I understand what you're saying and can land either way. However, I feel this code path that is used is a precompile that is only used for tests, having an automatic way of the developer refreshing this data without having to also manage a list of files to me is better DX for us as we make changes in the future. Also I intend that this code only runs when a developer makes a significant change to the state of our contracts and will commit the output. The magic would only run on the developer side and they would commit the output

scripts/python/build_selector_library.py
snissn: done! df2a263f

Makefile
snissn: opened issue https://github.com/consensus-shipyard/ipc-monorepo/issues/493

scripts/python/build_selector_library.py
snissn: sounds good! The previous implementation that this replaces had dynamic discover magic and that actually is my guess to the cause for our ci problems. Updated to remove directory traversal and just a solid list of filenames now

snissn:
Opened new follow up issue for a CI integration on this, otherwise I think I have updated with all your feedback @raulk

raulk (approved):

raulk (approved):
🎉

PR #411, opened 2023-12-21
Split SubnetActor

dnkolegov:
This PR :

- Splits Subnet Actor manager
- Fixes Pausable contract


src/interfaces/ISubnetActor.sol
```diff
@@ -49,3 +48,53 @@ interface ISubnetActor {
     /// The reword includes the fixed reward for a relayer defined in the contract and `amount` of fees from the cross-messages.
     function distributeRewardToRelayers(uint256 height, uint256 amount, QuorumObjKind kind) external payable;
 }
+
+interface ISubnetActorXX {
```
raulk, line 52: This seems duplicated, @dnkolegov? Is this a merge snafu?

src/subnet/SubnetActorCheckpointingFacet.sol
```diff
@@ -0,0 +1,156 @@
...
+import {Validator, ValidatorSet} from "../structs/Subnet.sol";
+import {MultisignatureChecker} from "../lib/LibMultisignatureChecker.sol";
+import {ReentrancyGuard} from "../lib/LibReentrancyGuard.sol";
+import {SubnetActorModifiers} from "../lib/LibSubnetActorStorage.sol";
+import {LibValidatorSet, LibStaking} from "../lib/LibStaking.sol";
+import {EnumerableSet} from "openzeppelin-contracts/utils/structs/EnumerableSet.sol";
+import {LibSubnetActor} from "../lib/LibSubnetActor.sol";
+import {Pausable} from "../lib/LibPausable.sol";
+
+contract SubnetActorCheckpointingFacet is SubnetActorModifiers, ReentrancyGuard, Pausable {
```
raulk, line 16: Should we just drop the `SubnetActor` prefix from the contract name, since the scope is already implied by the import path? I find it too verbose.

src/subnet/SubnetActorPauseFacet.sol
```diff
@@ -0,0 +1,24 @@
...
+        _pause();
+    }
+
+    /// @notice Unpauses all contract functions with the `whenNotPaused` modifier.
+    function unpause() external {
+        LibDiamond.enforceIsContractOwner();
+        _unpause();
+    }
+
+    /// @notice Unpauses all contract functions with the `whenNotPaused` modifier.
```
raulk, line 20: Copy-pasta.

src/subnet/SubnetActorRewardFacet.sol
```diff
@@ -0,0 +1,70 @@
...
+import {MethodNotAllowed} from "../errors/IPCErrors.sol";
+import {IRelayerRewardDistributor} from "../interfaces/ISubnetActor.sol";
+import {QuorumObjKind} from "../structs/Quorum.sol";
+import {Pausable} from "../lib/LibPausable.sol";
+import {ReentrancyGuard} from "../lib/LibReentrancyGuard.sol";
+import {SubnetActorModifiers} from "../lib/LibSubnetActorStorage.sol";
+import {LibStaking} from "../lib/LibStaking.sol";
+import {LibSubnetActor} from "../lib/LibSubnetActor.sol";
+
+contract SubnetActorRewardFacet is IRelayerRewardDistributor, SubnetActorModifiers, ReentrancyGuard, Pausable {
```
raulk, line 13: Could you explain the rationale for introducing a new interface here? (`IRelayerRewardDistributor`)

test/IntegrationTestBase.sol
```diff
@@ -366,17 +415,41 @@ contract IntegrationTestBase is Test, TestParams, TestRegistry, TestSubnetActor,
...
+                functionSelectors: saCutterSelectors
+            })
+        );
+
+        diamondCut[6] = (
             IDiamond.FacetCut({
                 facetAddress: address(louper),
                 action: IDiamond.FacetCutAction.Add,
-                functionSelectors: gwLoupeSelectors
+                functionSelectors: saLouperSelectors
```
raulk, line 452: Are these selectors and the cutter selectors `sa` or `gw` dependent? They are general-purpose, right?

raulk (approved):
Generally LGTM, my comments are not blockers (except for the apparent merge snafu) and we can address them in the monorepo.

src/interfaces/ISubnetActor.sol
dnkolegov: Sorry! will fix

src/subnet/SubnetActorCheckpointingFacet.sol
dnkolegov: Do you mean rename to Checkpointingfacet ?

test/IntegrationTestBase.sol
dnkolegov: That's correct, but that was incorrect either way due to `gw ` prefix

src/lib/LibPausable.sol
```diff
@@ -16,7 +16,7 @@ abstract contract Pausable {
     event Paused(address account);
 
     /**
-     * @dev Emitted when the pause is lifted by `account`.
+     * @dev Emitted when the unpause is lifted by `account`.
```
raulk, line 19: This description was correct before. `lifted` means "removed".

src/lib/LibPausable.sol
```diff
@@ -86,9 +99,9 @@ abstract contract Pausable {
      *
      * - The contract must be paused.
      */
-    function _unpause() internal {
-        _requirePaused();
+    function _unpause() internal  {
         PausableStorage storage s = pausableStorage();
```
raulk, line 103: What difference does this make?

src/lib/LibPausable.sol
dnkolegov: Probably, that was not my text. Fixed

src/lib/LibPausable.sol
dnkolegov: Use the same style across the codebase: first, get the storage, then do something.

src/subnet/SubnetActorCheckpointingFacet.sol
raulk: Yep

src/lib/LibPausable.sol
raulk: IMO this general pattern is unsafe. You should only fetch storage **_immediately_** _before_ using it. Otherwise you could be doing a dirty write if you call other functions in between that end up touching that storage. For example, what if `_requirePaused()` wrote to storage? (it doesn't in this case, but in the future it could). The previous was safer.

dnkolegov: unchanged

src/subnet/SubnetActorCheckpointingFacet.sol
dnkolegov: Then we will have two contracts with the same name: router/Checkpointingfacet and subnet/CheckpointinfFacet

